PR #6, opened 2026-07-02
fix: surface daemon command errors without respawning the daemon

onevcat:
## Summary

The daemon client's fast path treated **every** `classify` failure as a stale-daemon transport failure — including `ok=false` envelopes a perfectly healthy daemon answered with. For any failing command (element not found, bad selector, …) the client would:

1. delete the live daemon's socket + pidfile,
2. spawn a fresh daemon,
3. **re-send the request, executing the command a second time**.

Consequences: side-effecting commands (`type`, `tap --wait-timeout`) could double their effects, a healthy daemon was orphaned until its 600 s idle timeout (still holding FBSimulatorControl/HID connections), and every failed command paid a full respawn before the error reached the user.

## Fix

- `DaemonClientError.remote` is now rethrown on the fast path: the daemon answered, so its response **is** the command's outcome. Only transport-level failures (connect/write/read, empty/malformed response) still trigger stale-file cleanup + respawn.
- The retry-once behaviour for `transient_booting` that `DaemonClient`'s header has always documented is now actually implemented: one bounded re-send against the **same** daemon after a short delay (post-`simctl boot` accessibility-readiness gap). A second `transient_booting` propagates.

New test seams on `invoke`: `baseDirectory` (isolated daemon tree) and `transientRetryDelay`, both defaulted so production call sites are unchanged.

## Tests

`Tests/DaemonClientInvokeErrorRoutingTests.swift` (TDD, red → green), driving a real in-process `DaemonServer` with an injected command parser:

- remote error surfaces verbatim; socket/pidfile survive; command executed exactly once
- `transient_booting` retried once against the same daemon, then succeeds
- persistent `transient_booting` propagates after exactly one retry

Full suite: 558 tests / 108 suites pass.

## Verification (live simulator)

A failing `tap --label` now completes in **0.77 s (was 1.8 s)**, executes once, and leaves the daemon pid unchanged — previously the trace showed `Removing stale files` → `spawning fresh daemon` → duplicated execution, and the orphaned daemon stayed alive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Daemon client now retries a command once against the same daemon when the simulator reports the post-boot `transient_booting` readiness gap, matching the long-documented behaviour.
+
+### Fixed
+
+- Daemon client no longer tears down a healthy daemon and re-executes the command when the daemon answers with a command-level error (element not found, etc.). Failed commands now surface immediately instead of paying a full daemon respawn, and side-effecting commands are no longer executed twice.
+
 ## [0.9.0] - 2026-06-29
 
 Initial public release.
--- a/Sources/SimUseCore/Daemon/DaemonClient.swift
+++ b/Sources/SimUseCore/Daemon/DaemonClient.swift
@@ -63,6 +63,13 @@ public enum DaemonClient {
                     transientRetryDelay: transientRetryDelay
                 )
             } catch {
+                // Cooperative cancellation is the caller's signal, not
+                // a stale daemon — the retry delay below is a
+                // suspension point, so it can surface here. Rethrow
+                // before any cleanup/respawn.
+                if error is CancellationError {
+                    throw error
+                }
                 // A `remote` error means the daemon is alive and
                 // answered `ok=false` — that IS the command's outcome.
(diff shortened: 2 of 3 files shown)


onevcat:
Review follow-up (9be8d72), addressing two findings from agent review:

1. **CancellationError misread as transport failure (Medium)** — the retry delay added this PR's first suspension point on the fast path, so a cancelled task could fall into the stale-daemon cleanup (delete files → respawn → re-execute). `CancellationError` is now rethrown before any cleanup, and the retry re-checks cancellation before re-sending. Regression test: cancel lands inside the retry sleep → `CancellationError` propagates, daemon files survive, command executed exactly once (test runtime dropped 30 s → 4.5 s, confirming the cancel now actually interrupts the delay).
2. **Non-finite `transientRetryDelay` traps (Low)** — public parameter; `.infinity`/NaN hit a fatal `UInt64` conversion. Non-finite values now skip the pause (immediate retry), finite values are capped at 60 s so the nanosecond conversion cannot overflow. Regression test: `.infinity` completes the retry without trapping.

Branch suite: 560 tests / 108 suites pass.

onevcat:
Follow-up 82c2fb8: dropped the redundant `try` on the cancellation test's `withParser` call (its closure handles every error itself, so `rethrows` resolves to non-throwing) — clears the "no calls to throwing functions occur within 'try' expression" warning. Verified: recompiling the test target and daemon sources now emits zero warnings; suite still 5/5.

PR #7, opened 2026-07-02
fix: guard daemon file ownership on cleanup and validate the base directory

onevcat:
## Summary

Two daemon filesystem-lifecycle fixes.

**Orphan cleanup deletes the successor's files.** `DaemonServer.cleanup()` removed its socket/pidfile unconditionally. When a successor daemon takes the paths over (`listen()` unlinks + rebinds the socket; the pidfile is overwritten), an orphaned predecessor shutting down later — idle timeout or SIGTERM — deleted the **live** daemon's files. The next client then saw no daemon and spawned a third one, chaining invisible orphans that each pin FBSimulatorControl state. Observed live during review: an orphan's idle-timeout wiped the pidfile of the daemon actively serving the UDID (visible in `/tmp/sim-use-<uid>/<udid>.log`).

**Base directory trusts whatever pre-exists.** `DaemonPaths.ensureBaseDirectory()` applied its 0700 mode only on first creation and silently accepted a pre-existing path at world-writable `/tmp/sim-use-<uid>` — including a directory or symlink pre-planted by another local user, who could then swap sockets and pidfiles underneath us (the doc comment claimed 0700 protection the implementation didn't guarantee).

## Fix

- `cleanup()` leaves the socket/pidfile alone when the pidfile holds another process's pid; missing/unparseable pidfile keeps the old stale-cleanup behaviour.
- `ensureBaseDirectory()` lstat-validates on every run: symlinks/non-directories and foreign owners are rejected with an actionable `DaemonPathsError`, loose group/other permission bits are tightened back to 0700.

## Tests (TDD, red → green)

- `Tests/DaemonServerCleanupOwnershipTests.swift` — starts a real `DaemonServer`, simulates a path takeover (foreign pidfile + replaced socket file), stops the predecessor over a pre-takeover connection, and asserts the successor's files survive.
- `DaemonPathsTests` — new suite: fresh creation is 0700 (pin), pre-existing 0777 dir is tightened to 0700, symlink at the base path is rejected.

Full suite: 559 tests / 109 suites pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Daemon shutdown no longer deletes the socket/pidfile of a successor daemon that took the paths over, which previously chained invisible orphan daemons.
+- Daemon base directory under `/tmp` is now validated on every run (symlinks, foreign owners rejected; loose permissions tightened to 0700) instead of trusting whatever was pre-created there.
+
 ## [0.9.0] - 2026-06-29
 
 Initial public release.
--- a/Sources/SimUse/Commands/Daemon.swift
+++ b/Sources/SimUse/Commands/Daemon.swift
@@ -172,7 +172,7 @@ struct Daemon: AsyncParsableCommand {
         }
 
         func execute() async throws -> ExecutionResult {
-            let targets = resolveTargets()
+            let targets = try resolveTargets()
             var entries: [StopEntry] = []
             for target in targets {
                 entries.append(await stopOne(target))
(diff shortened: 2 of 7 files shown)


onevcat:
Review follow-up (bbde594), addressing two findings from agent review:

1. **Base-directory gate bypassed on management paths (Medium)** — the lstat/owner/mode checks only ran via `ensureBaseDirectory` on spawn/connect. `daemon stop --all` / `daemon status` / the device-resolver fast path enumerate the default tree directly, and stop's fallback SIGTERMs the pids it finds — so a pre-planted `/tmp/sim-use-<uid>` with forged pidfiles could steer a victim into signalling arbitrary processes they own. The checks are extracted into a shared `DaemonPaths.validateBaseDirectory(at:)`; `enumerateLiveDaemons` now throws on a failing tree (absent tree still yields `[]`), and — one step beyond the finding — `stop --device <id>`'s single-target branch validates too, since it trusts `filesystemLiveness` on the same default tree. Tests: symlink/regular-file at the base path → throw; loose-perm own dir → tightened to 0700, then enumerated.
2. **Cleanup ownership honoured for dead owners (Low)** — `cleanup()` deferred to any foreign pid in the pidfile; a dead owner's files are stale garbage (stranded until the next client's liveness probe, and a recycled pid could later fake a live daemon). Ownership is now honoured only while the owner is alive. Tests: takeover by a **live** foreign process (spawned `/bin/sleep` child) → files survive; takeover by a dead pid → files cleaned on shutdown.

One nuance vs the review text: a dead-owner residue never produced a failed client request — `filesystemLiveness` cleans dead-pid files before any send — the real costs were the stale residue itself and the pid-recycle edge. The fix is the same either way.

Branch suite: 563 tests / 110 suites pass.

PR #9, opened 2026-07-02
fix: iOS HID hot paths — one type session, record-video frame pacing

onevcat:
## Summary

Two iOS-backend hot-path fixes, one commit each.

**`type` re-initialised FBSimulatorControl per character.** `IOSSimTypeCommand` called the per-UDID `performHIDEvent` overload for every HID event, and that overload runs `makeSession` — private-framework load plus a fresh `FBSimulatorControl` set construction — on each call. Only the HID connection inside was cached, so typing paid the whole setup cost once per character. The command now builds a single session up front and feeds every event through it, matching what the batch runner already does.

**`record-video` hot-spins on persistent decode failure.** The decode-failure guard used `continue`, skipping the frame-pacing sleep at the bottom of the capture loop — if `makeCGImage` kept returning nil the loop spun at full speed. The failure path now falls through to the same pacing as a successful frame (consistent with how `stream-video`'s error path already behaves).

## Tests / verification

Neither path is reachable from unit tests without a live simulator; both changes are control-flow-local.

- Full suite: 555 tests / 107 suites pass (baseline unchanged — no behavioural surface in unit scope).
- Live simulator: `type "hello sim-use"` lands correctly into Spotlight through the daemon path (1.5 s end-to-end for 13 chars).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- `ios type` builds one HID session for the whole string instead of re-initialising FBSimulatorControl per character.
+
+### Fixed
+
+- `record-video` no longer hot-spins without frame pacing when screenshot frames persistently fail to decode.
+
 ## [0.9.0] - 2026-06-29
 
 Initial public release.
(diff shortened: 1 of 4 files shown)


onevcat:
Review follow-up (42f38b2): fixed the empty-input regression the session-reuse change introduced.

Sharing one HID session moved `makeSession` above the event loop, so `type ""` (zero HID events) now paid framework + FBSimulatorControl-set init and could fail with "not found in set" against a non-booted device — previously it skipped the loop entirely and was a silent no-op that an agent's `type "$VAR"` with an empty variable relies on. Added an early return before session creation when there are no events.

Regression test (`TypeForwarderTests`, red → green): a parsed `type ""` against an iOS-shaped UDID that matches no simulator — reaching `makeSession` throws "not found in set" (confirmed the red), now succeeds as a no-op.

Branch suite: 556 tests / 107 suites pass.

PR #11, opened 2026-07-02
fix: drop keyboard-state's vestigial run() override

onevcat:
## Summary (D1 from the post-review discussion)

`KeyboardState` carried a hand-rolled `run()` override from a **pre-0.5.x** design where `hidden` exited 1. That contract was later changed — both `soft` and `hidden` now exit 0, which falls out of `execute()` returning a result (`visible: true/false`) instead of throwing — but the override was left behind. Its own comment had gone self-contradictory ("overrides run() … Both `soft` and `hidden` now exit 0"), and the test file still documented the removed exit-1 behaviour.

With its original purpose gone, the override only diverged from the protocol default in ways nobody intended:

- **No daemon routing** — every `keyboard-state` call paid full framework + `FBSimulatorControl` init instead of amortising through the per-UDID daemon the way `describe-ui` does. The two entry points to the same verb even disagreed: `sim-use ios keyboard-state` (no override) routed through the daemon; top-level `sim-use keyboard-state` did not.
- **No crash-advisory banner** — the process-liveness banner other verbs print was silently suppressed.
- **No `Hint:` line** — hint-bearing errors lost their actionable suggestion on the text path.

None of these was a feature. Removing the override adopts the default `run()` and restores all three, while keeping the exit-code contract byte-for-byte: `execute()` returns a result for both keyboard states (exit 0) and only throws on genuine failure (exit 1).

## Tests

TDD, red → green. New e2e suite `KeyboardState — routes through the per-UDID daemon` pins the restored routing (after a successful call, `daemon status` lists a daemon for the UDID). It **fails** with the override in place (`daemons:[]`, inline execution) and **passes** after removal. The existing resolve-wiring and UDID-echo suites still pass under the default `run()`; their stale "overrides run()" comments are corrected.

- `make test` (unit): 570 tests / 112 suites pass.
- E2E with a booted simulator (`SIM_USE_E2E=1 SIMULATOR_UDID=…`): all four KeyboardState suites pass.
- Live spot-checks: `hidden` → exit 0, `--json` → exit 0, bogus UDID → exit 1 with a `Hint:` line now present, `daemon status` shows the spawned daemon after a call (empty before).

> Branches off current `main`; no CHANGELOG conflict expected with anything outstanding.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bridge `/a11y_tree_full` no longer reads the active root's `windowId` after the node was recycled, which silently dropped popup/dialog secondary windows on Android 11–12.
 - Bridge `/keyboard/input` no longer leaks the borrowed root `AccessibilityNodeInfo` on every call.
 - `record-video` no longer hot-spins without frame pacing when screenshot frames persistently fail to decode.
+- `keyboard-state` now routes through the per-UDID daemon like every other verb (amortised init) and surfaces crash advisories and error `Hint:` lines; a vestigial `run()` override had silently opted it out. The `soft`/`hidden`/failure exit codes are unchanged.
 
 ## [0.9.0] - 2026-06-29
 
--- a/Sources/SimUse/Commands/KeyboardState.swift
+++ b/Sources/SimUse/Commands/KeyboardState.swift
@@ -11,10 +11,10 @@ import iOSSimBackend
 /// Simulator UDIDs, `AndroidKeyboardStateCommand.performKeyboardState`
 /// for adb serials).
 ///
-/// Overrides `run()` so the JSON envelope shape (`{ok, data}` /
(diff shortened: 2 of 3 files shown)


onevcat:
Follow-up b585a3e, addressing the test-robustness finding: the daemon-routing e2e test now asserts `daemon status --json` does **not** list the UDID *before* `keyboard-state` runs, right after the `daemon stop`. Previously it only checked post-call presence, so a stale/pre-existing same-UDID daemon (stop failed, or the box already had one) could satisfy the check even under the inline-execution regression — a false pass. With the explicit precondition, "empty before → present after" is unambiguously attributable to keyboard-state spawning it, and a failed stop now fails the test loudly instead of masking a regression.

Verified with a booted simulator (`SIM_USE_E2E=1 SIMULATOR_UDID=…`): the routing suite passes, precondition holds after stop. Unit suite unaffected (570/112).

Good call — the assumption that CI never has a residual same-UDID daemon was exactly the fragile part; the precondition removes the dependency on it.

PR #16, opened 2026-07-02
Document frame units in SKILL.md; correct keep-alive comment

onevcat:
## Summary

Deferred review items **D7** and **D5**, both resolved as "document reality, no behavior change" per owner decision.

### D7 — frame units

`Outline.Frame` values are platform-native (iOS **points**, Android **pixels**). Verification showed the describe-ui JSON envelope already exposes a machine-readable `platform` field, so consumers can disambiguate without any wire change. Added one line to `skills/sim-use/SKILL.md` §Observe telling agents to key off `platform` before doing cross-platform coordinate math.

### D5 — keep-alive comment

`BridgeClient.swift` claimed the URLSession reuses a TCP connection across requests, but the bridge's `HttpServer.kt` sends `Connection: close` on every response — reuse never happens. The comment now describes reality (fresh loopback connect per call, cheap through `adb forward`; server-side keep-alive intentionally not implemented).

## Testing

- `make build` OK.
- `make test` — 579 tests / 115 suites pass.

### DIFF
--- a/Sources/AndroidBackend/Bridge/BridgeClient.swift
+++ b/Sources/AndroidBackend/Bridge/BridgeClient.swift
@@ -69,9 +69,11 @@ public final class BridgeClient: @unchecked Sendable {
         let config = URLSessionConfiguration.ephemeral
         config.timeoutIntervalForRequest = readTimeout
         config.timeoutIntervalForResource = readTimeout
-        // Reuse a single TCP connection across requests — `adb forward`
-        // is a stream-multiplexer and benefits massively from keep-alive
-        // (saves the connect handshake on every call).
+        // No connection reuse in practice: the bridge answers every
+        // request with `Connection: close` (HttpServer.kt), so each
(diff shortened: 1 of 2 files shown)


onevcat:
## Review

- **Keep-alive comment (D5)**: confirmed `HttpServer.kt`'s single response writer (`writeResponse`) unconditionally sends `Connection: close`, so the old "reuse a single TCP connection" comment was indeed wrong. The new wording (fresh loopback connect per call; the connection cap only matters if the server ever gains keep-alive) matches the code.
- **Frame units (D7)**: the describe-ui JSON envelope does expose `platform` (`DescribeUI.swift`), and the key names cited in the new SKILL.md line (`entries[].frame`, `screen`) match the actual output struct. Documenting platform-native units (iOS points / Android pixels) with no wire change is the right call — the disambiguator was already there.
- Zero behavior change; comment/docs only.

Re-ran locally: `make build` OK, `make test` — 579 tests / 115 suites passed.

PR #17, opened 2026-07-02
Fix Viewer API false success on non-zero subprocess exit

onevcat:
## Summary

Fixes deferred review item D15: the Viewer's API handlers (`ViewerAPIHandlers.swift`) shelled out to the sim-use binary and trusted only the `{ok:false}` JSON envelope for failure detection — the child's exit status was discarded for any normal exit. Two failure modes resulted:

1. Non-zero exit + stdout parsing as a JSON object without an `ok` field → the handler fell through to its success path and returned `200 {ok:true}` — false success.
2. Non-zero exit + empty/non-JSON stdout → the caller saw a useless `NSCocoaErrorDomain 3840` parse-error string while the real error text on stderr was dropped.

## Changes

- `run()` now returns `(stdout: Data, stderr: Data, status: Int32)` instead of swallowing the exit status on `.exit` terminations (signal kills still throw as before).
- A shared `failureResponse(envelope:result:)` helper applies one policy across `devices`, `snapshot`, and `tap`:
  - envelope with `ok == false` → forwarded verbatim as 502, regardless of exit code (CLI domain errors legitimately exit non-zero WITH an envelope);
  - no usable `ok` field AND non-zero exit → 502 `{ok:false, error: <trimmed stderr, falling back to stdout, falling back to "sim-use exited with status N">}`;
  - zero exit → existing behavior, including the "bad JSON" 502 guards in `devices`/`snapshot`.

## Testing

- New `Tests/ViewerAPIHandlersTests.swift` (9 tests, TDD-first — 6 failed against the old code): points `ViewerAPIHandlers(executable:)` at per-test `/bin/sh` scripts with scripted stdout/stderr/exit codes to cover the policy matrix, including the false-success regression (`exit 2` + `{"unexpected":"shape"}`) and the stderr surfacing (`exit 3` + `boom` on stderr) for tap/devices/snapshot.
- `make build` + full `make test`: 588 tests in 116 suites passed (baseline 579/115).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Daemon no longer shuts itself down in the middle of a request that runs longer than the idle timeout (e.g. `tap --wait-timeout` beyond the timeout, or a long `batch`). The idle timer now defers shutdown while a request is in flight.
 - Daemon client no longer respawns and resends a command when the daemon drops the connection *after* receiving the request (a possible mid-execution crash). Such ambiguous outcomes now surface a dedicated error with a hint to re-observe the screen before retrying, so a side-effecting verb (tap/type/swipe) is never silently applied twice. Pre-delivery failures (connect/write) still respawn as before.
 - `keyboard-state` now routes through the per-UDID daemon like every other verb (amortised init) and surfaces crash advisories and error `Hint:` lines; a vestigial `run()` override had silently opted it out. The `soft`/`hidden`/failure exit codes are unchanged.
+- Viewer API no longer reports success when the underlying sim-use invocation exits non-zero without a parseable JSON envelope; the subprocess's stderr is now surfaced in the error response instead of a generic JSON-parse failure.
 
 ## [0.9.0] - 2026-06-29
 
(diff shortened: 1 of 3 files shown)


onevcat:
## Review

- **Bug confirmed real**: the old `tap` path fell through to `200 {ok:true}` whenever stdout parsed as a JSON object lacking `ok` — regardless of exit code — and `devices`/`snapshot` masked non-zero-exit failures behind an `NSCocoaErrorDomain 3840` parse error while dropping stderr. The new `failureResponse` policy fixes both.
- **Policy is sound**: envelope `ok` stays authoritative (domain errors that exit non-zero *with* an envelope are still forwarded verbatim), and exit status only kicks in when no usable envelope exists. One deliberate semantic change worth noting: `tap` with exit 0 + unparseable stdout now returns 200 (previously a 500 via the parse-error catch). That's the more correct reading — exit 0 is the CLI's success contract and the tap handler never consumes the envelope body — and it avoids the SPA re-tapping a tap that actually landed.
- **Tests**: the 9-test matrix covers the regression (`exit 2` + ok-less JSON), stderr surfacing, the stdout/status-message fallbacks, and both preserved success paths. I ran the new suite 10× in isolation — stable every time.
- **Flake note (pre-existing, unrelated)**: one of my three full-suite runs failed on `AdbRunnerTests.testRunFastChildHasLowLatency` (`0.394s !< 0.3s`) — the known load-sensitive latency assertion that also flakes on `main`. Not introduced by this PR.

Re-ran locally: `make build` OK, `make test` — 588 tests / 116 suites passed (2 of 3 runs; third failed only on the pre-existing AdbRunner flake).

PR #18, opened 2026-07-02
fix: record-video stop watchdog no longer exits 0 on finalization overrun

onevcat:
## Problem (deferred review item D13)

The `record-video` SIGINT/SIGTERM handler armed a watchdog:

```swift
DispatchQueue.global().asyncAfter(deadline: .now() + 1.5) {
    if !recordingFinished.isCancelled() { _exit(0) }
}
```

The stop path breaks the frame loop, then awaits `recorder.finish()` (`AVAssetWriter.finishWriting`), and `recordingFinished.cancel()` is only set after `recordVideo` returns. If `finishWriting` takes longer than 1.5 s (large file, slow disk), `_exit(0)` fires mid-finalization: the MP4's moov atom is never written — a truncated, possibly unplayable file with **exit status 0**.

The identical pattern existed at both watchdog sites: `IOSSimRecordVideoCommand.execute()` (iOS) and `RecordVideo.executeAndroid()` (Android branch of the top-level forwarder).

## Fix

Keep the watchdog — it still bounds a genuinely hung `finishWriting` — but make it honest:

- Grace window widened 1.5 s → 3 s.
- On overrun it now writes `warning: video finalization did not complete within 3s — output file may be truncated or unplayable` to stderr and exits **70** (`EX_SOFTWARE`) instead of 0.

The policy (grace period, exit code, message) plus the arm routine are extracted into `RecordingFinishWatchdog` in `VideoCommandSupport.swift`, shared by both call sites so they can't drift again.

Exit code 70 does not collide with existing exit paths: the CLI otherwise produces 0 (success), 1 (runtime error via `CLIError`), and 64 (ArgumentParser usage errors). `record-video` is `daemonBypass = true`, so stderr at the watchdog site is the invoking terminal/pipe, not a daemon log.

## Testing

- `make build` green; `make test` green — 582 tests / 116 suites (baseline 579 / 115 + the 3 new tests below).
- New `Tests/RecordingFinishWatchdogTests.swift` pins the watchdog contract: exit code 70 (and explicitly `!= 0`), 3 s grace window, warning message shape. The `_exit` side effect itself fires from a detached dispatch closure and is not unit-testable in-process.
- Manual verification is a code-path walkthrough (no booted simulator available): signal → `cancellationFlag.cancel()` + `RecordingFinishWatchdog.arm(...)` → frame loop breaks → `recorder.finish()` → on return (success or error) `recordingFinished.cancel()` disarms the watchdog before it can fire; only an overrunning/hung `finish()` reaches the warning + `_exit(70)` branch. The fast path (finalization under 3 s) is behaviorally unchanged. The existing E2E regression test for issue #35 (`RecordVideoTests.recordVideoShortGraceSIGTERM`) still covers the signal → valid-mp4 path when run with a simulator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Daemon no longer shuts itself down in the middle of a request that runs longer than the idle timeout (e.g. `tap --wait-timeout` beyond the timeout, or a long `batch`). The idle timer now defers shutdown while a request is in flight.
 - Daemon client no longer respawns and resends a command when the daemon drops the connection *after* receiving the request (a possible mid-execution crash). Such ambiguous outcomes now surface a dedicated error with a hint to re-observe the screen before retrying, so a side-effecting verb (tap/type/swipe) is never silently applied twice. Pre-delivery failures (connect/write) still respawn as before.
 - `keyboard-state` now routes through the per-UDID daemon like every other verb (amortised init) and surfaces crash advisories and error `Hint:` lines; a vestigial `run()` override had silently opted it out. The `soft`/`hidden`/failure exit codes are unchanged.
+- `record-video`'s stop watchdog no longer exits 0 when video finalization overruns its grace window, which could report success for a truncated/unplayable MP4. It now warns on stderr and exits 70 (`EX_SOFTWARE`), and the grace window is 3 s (was 1.5 s).
 
 ## [0.9.0] - 2026-06-29
 
--- a/Sources/SimUse/Commands/RecordVideo.swift
+++ b/Sources/SimUse/Commands/RecordVideo.swift
@@ -112,11 +112,7 @@ struct RecordVideo: SimUseExecutableCommand {
         let recordingFinished = CancellationFlag()
         let signalObserver = SignalObserver(signals: [SIGINT, SIGTERM]) {
             cancellationFlag.cancel()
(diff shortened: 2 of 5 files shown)


onevcat:
## Review

- **Bug confirmed real**: both removed sites armed `_exit(0)` 1.5 s after the stop signal with `recordingFinished` only cancelled *after* `recorder.finish()` returns — so a slow `finishWriting` produced a moov-less MP4 with success status. Truncated-file-as-exit-0 is exactly the failure a supervisor can't detect.
- **Coverage is complete**: verified there is no third watchdog site — `AndroidBackend` registers no record verb, so Android recording is only reachable through the top-level forwarder's `executeAndroid()`. Both real sites now share `RecordingFinishWatchdog`, so the policy can't drift again.
- **Disarm paths are correct**: `recordingFinished.cancel()` fires on both the success and the error path at both call sites, so a throwing `finish()` still exits via `CLIError` (status 1) rather than lingering into the watchdog.
- **Claims verified**: both commands are `daemonBypass = true` (stderr goes to the invoking terminal), and a repo-wide grep confirms no other exit path uses 70 — no collision with 0/1/64.
- **Safety**: `SignalObserver` is `DispatchSourceSignal`-based, so the handler runs on a dispatch queue — `asyncAfter` and the stderr write in the watchdog closure are safe (not an async-signal context).
- The new tests only pin constants/message shape, but the `_exit`-from-detached-closure side effect genuinely isn't unit-testable in-process; the pinning at least locks the "never exit 0 on overrun" contract against regression.

Re-ran locally: `make build` OK, `make test` — 582 tests / 116 suites passed.

PR #19, opened 2026-07-02
Fail record-video loudly when the asset writer stalls

onevcat:
## Summary

`H264StreamRecorder.append` busy-waited on `input.isReadyForMoreMediaData` with no timeout. `append` is synchronous and called from the recording loops' async context, so a wedged `AVAssetWriter` (disk pressure, encoder stall) parked a Swift-concurrency cooperative-pool thread and hung the recording task forever with no escape.

## Changes

- Extract the wait into `H264StreamRecorder.waitUntilReady(isReady:timeout:pollInterval:now:sleep:)` with a generous 10 s deadline. With `expectsMediaDataInRealTime = true` a healthy writer drains within milliseconds, so hitting the deadline means the writer is wedged, not merely busy.
- On timeout the helper throws a dedicated `VideoWriterStallError` (`LocalizedError`) with a descriptive message. This is an intentional behavior change on the failure path: a stalled writer now fails the recording loudly instead of hanging.
- Both frame loops (iOS `IOSSimRecordVideoCommand.recordVideo` and the Android branch in `RecordVideo`) rethrow `VideoWriterStallError` past their transient per-frame `catch`, so the command exits non-zero with the message instead of re-logging the stall once per timeout forever. Transient capture errors keep their existing log-and-continue behavior.

## Testing

The timeout policy is a pure function with injectable `now`/`sleep` hooks, so it is unit-tested without AVFoundation or real sleeping (`Tests/VideoWriterStallTimeoutTests.swift`):

- never-ready input throws `VideoWriterStallError` exactly at the deadline (fake clock, instant);
- ready-immediately returns without sleeping;
- ready after N polls returns after exactly N sleep calls;
- production timeout constant is 10 s; stall error message is descriptive.

Plus a short real-clock AVFoundation integration test: an `AVAssetWriterInput` whose writer never started reports `isReadyForMoreMediaData == false` forever, and trips the helper with a 50 ms timeout (runs headless in ~60 ms).

`make build` and `make test` green: 585 Swift Testing tests / 116 suites + 302 XCTest tests, 0 failures. (`AdbRunnerTests.testRunFastChildHasLowLatency` flaked once under concurrent build load on the *unmodified* baseline too; it passes in isolation and in the final clean run.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bridge `/a11y_tree_full` no longer reads the active root's `windowId` after the node was recycled, which silently dropped popup/dialog secondary windows on Android 11–12.
 - Bridge `/keyboard/input` no longer leaks the borrowed root `AccessibilityNodeInfo` on every call.
 - `record-video` no longer hot-spins without frame pacing when screenshot frames persistently fail to decode.
+- `record-video` no longer hangs forever when AVAssetWriter stops accepting frames; a stalled writer now fails the recording with an explicit error after 10 s.
 - Daemon no longer shuts itself down in the middle of a request that runs longer than the idle timeout (e.g. `tap --wait-timeout` beyond the timeout, or a long `batch`). The idle timer now defers shutdown while a request is in flight.
 - Daemon client no longer respawns and resends a command when the daemon drops the connection *after* receiving the request (a possible mid-execution crash). Such ambiguous outcomes now surface a dedicated error with a hint to re-observe the screen before retrying, so a side-effecting verb (tap/type/swipe) is never silently applied twice. Pre-delivery failures (connect/write) still respawn as before.
 - `keyboard-state` now routes through the per-UDID daemon like every other verb (amortised init) and surfaces crash advisories and error `Hint:` lines; a vestigial `run()` override had silently opted it out. The `soft`/`hidden`/failure exit codes are unchanged.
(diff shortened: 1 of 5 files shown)


onevcat:
## Review

- **Bug confirmed real**: the removed code busy-waited `while !input.isReadyForMoreMediaData` with no exit condition — a wedged writer parked a cooperative-pool thread and hung the recording forever with no diagnostic. The 10 s deadline is the right order of magnitude given `expectsMediaDataInRealTime` drains in milliseconds when healthy.
- **Call-site coverage verified**: `append(image:presentationTime:)` has exactly 4 call sites, all in the two record-video paths. The two in-loop sites get the specific `VideoWriterStallError` rethrow (correctly ordered before the generic log-and-continue catch); the two initial-frame sites sit outside the loops so the error already propagated loudly. No other consumer (stream-video uses a separate pipeline) picks up the new throw unexpectedly.
- **Skipping `finish()` on stall is correct**: both loops place `recorder.finish()` after the loop, so a stall aborts without calling it — deliberate and right, since a wedged writer's `finishWriting` would just hang again and nothing bounds it on the non-signal path.
- **Tests are well-designed**: the injectable `now`/`sleep` hooks make the deadline policy deterministic (including the exactly-representable 0.25 s poll trick), and the unstarted-`AVAssetWriterInput` integration test exercises the real `isReadyForMoreMediaData == false` shape headless. Ran the suite in isolation — 6/6 instant and stable.
- Merge note: touches the same three files as #18 but in disjoint regions — they merge cleanly in either order.

Re-ran locally: `make build` OK, `make test` — 585 tests / 116 suites passed.

PR #20, opened 2026-07-02
Fix `ios batch --ax-cache` being a complete no-op

onevcat:
## Problem

`sim-use ios batch --ax-cache perBatch|perStep|none` was a complete no-op:

- `BatchContext.accessibilityRoots(logger:forceRefresh:)` had **zero callers** — the only place `axCachePolicy` was read was inside that dead method, whose `.perStep` and `.none` case bodies were byte-identical anyway, and `forceRefresh` was never passed `true`.
- Selector-based tap steps resolved through `AccessibilityPoller.resolveWithPolling`, which called `AccessibilityFetcher.fetchAccessibilityElements` directly — a fresh full-tree fetch on every step, cache never consulted.

So the documented default (`perBatch` "reuses one AX snapshot for the whole run") silently behaved like `none`, paying a full AX tree fetch per selector step.

## Fix

Wire the cache in with the semantics the docs and the E2E tests already promise:

| Policy | Behaviour |
|---|---|
| `perBatch` (default) | One snapshot fetched at first use, reused across all steps. `forceRefresh` refetches **and replaces** the cache. |
| `perStep` | Snapshot cached within the current step only; the step loop calls the new `BatchContext.beginStep()` at each boundary to drop it. |
| `none` | Fetches on every resolution, never caches (differs from `perStep` only when one step resolves multiple selectors). |
| `--wait-timeout` polling | Initial attempt respects the cache; every poll tick passes `forceRefresh: true`, so delayed elements are found under any policy and the refreshed snapshot propagates to later `perBatch` steps. |

Plumbing:

- `AccessibilityPoller.resolveWithPolling` gains an optional `rootsProvider: (_ forceRefresh: Bool) async throws -> [AccessibilityElement]`. Batch tap resolution passes the `BatchContext` cache through it; non-batch callers (`ios tap`, `paste --via-menu`) pass nothing and keep the historical fetch-every-time behaviour.
- `BatchContext` gains an injectable `fetchElements` fetcher (production default unchanged) so the policy semantics are unit-testable without a booted simulator.
- `--ax-cache` help text, README, and `skills/sim-use/references/batch-reference.md` clarified to match.

## Tests

New `Tests/BatchAXCacheTests.swift` (10 tests, fetch-counting fake fetcher, no simulator needed):

- `BatchContext`: perBatch caches across `beginStep()` (stale snapshot served, 1 fetch); `forceRefresh` refetches + updates cache; perStep clears at step boundary; `none` always fetches.
- Poller provider contract: called with `(false)` then `(true)` across a poll tick; single `(false)` attempt when `--wait-timeout 0`.
- Parser-level integration through `BatchStepParser.parseStepTokens`: two tap-by-label steps → 1 fetch under perBatch, 2 under perStep; unit mirrors of the E2E stale-snapshot-fails and delayed-element-found scenarios.

`make test`: 589 tests / 118 suites green (baseline before this change: 579 / 115).

Note: the SIM_USE_E2E-gated `BatchTests` (`perBatchCacheCanFailOnStateChange`, `perStepCacheHandlesStateChange`, `waitTimeoutFindsDelayedElement`) pin these exact semantics against a live UI; not run here — no booted simulator. The new unit tests mirror those scenarios with a fake fetcher.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `ios batch --ax-cache` was a complete no-op: the default `perBatch` never cached and every selector-based step refetched the AX tree. `perBatch` now resolves all steps against one snapshot, `perStep` refetches at each step, `none` never caches, and `--wait-timeout` poll ticks bypass the cache (updating it) so delayed elements are still found.
 - Daemon client no longer tears down a healthy daemon and re-executes the command when the daemon answers with a command-level error (element not found, etc.). Failed commands now surface immediately instead of paying a full daemon respawn, and side-effecting commands are no longer executed twice.
 - Daemon shutdown no longer deletes the socket/pidfile of a successor daemon that took the paths over, which previously chained invisible orphan daemons.
 - Daemon base directory under `/tmp` is now validated on every run (symlinks, foreign owners rejected; loose permissions tightened to 0700) instead of trusting whatever was pre-created there.
--- a/README.md
+++ b/README.md
@@ -272,7 +272,7 @@ Key semantics:
 - Exactly one step source per run: `--step`, `--file`, or `--stdin`.
 - Fail-fast by default; `--continue-on-error` switches to best-effort.
 - `--wait-timeout <seconds>` makes selector taps poll for the element to appear — primary mechanism for multi-screen flows.
-- `--ax-cache perBatch` (default) reuses one AX snapshot for the whole run; `--ax-cache perStep` refreshes between steps when the UI changes.
+- `--ax-cache perBatch` (default) reuses one AX snapshot for the whole run; `--ax-cache perStep` refreshes between steps when the UI changes; `--ax-cache none` disables snapshot reuse entirely. `--wait-timeout` polling always refetches.
 
 ### Screenshot
 
--- a/Sources/iOSSimBackend/A11y/AccessibilityPoller.swift
+++ b/Sources/iOSSimBackend/A11y/AccessibilityPoller.swift
@@ -3,6 +3,11 @@ import Foundation
 
 @MainActor
 public struct AccessibilityPoller {
+    /// Supplies AX roots for resolution. `forceRefresh: false` may serve
+    /// a cached snapshot; `true` (poll ticks) must fetch a fresh one so
+    /// delayed elements become visible.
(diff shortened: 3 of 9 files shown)


onevcat:
## Review

**Verdict: ✅ Approve — with one test-only fix pushed to this branch.**

### Findings

- **No-op claim verified**: on `main`, `BatchContext.accessibilityRoots` has zero callers (repo-wide grep) and the poller always called `AccessibilityFetcher` directly — `--ax-cache` was indeed parsed and ignored.
- **Coverage is complete**: the only selector-resolution path in batch is the tap conversion in `Command+BatchConvertible.swift`; non-batch callers (`ios tap` ×2, `paste --via-menu`) pass no `rootsProvider` and keep fetch-every-time. No other consumer inherits the cache by accident.
- **The behavior change is contract-backed, not a regression**: `perBatch` (default) now serving a stale snapshot to later steps *is* a user-visible change vs. the accidental fetch-every-step behavior — but the SIM_USE_E2E `BatchTests` have pinned exactly these semantics since the initial 0.9.0 release (`perBatchCacheCanFailOnStateChange` requires step 2 to *fail* after a state change), and README/`--wait-timeout` remain the documented path for multi-screen flows (poll ticks bypass and update the cache).

### Fix pushed (30a779b): E2E BatchTests were unrunnable

All 14 `BatchTests` still invoked the removed top-level `sim-use batch` (moved to `sim-use ios batch` in 0.5.x), so every test died on the namespace error before touching the UI — meaning the E2E contract this PR cites could not actually be executed. I fixed the 14 invocations (test-only change) and ran them against a live iPhone 17 Pro simulator:

- **12/14 pass, including all three cache-semantics tests** (`perBatchCacheCanFailOnStateChange`, `perStepCacheHandlesStateChange`, `waitTimeoutFindsDelayedElement`) — the new plumbing holds up against real UI, not just the fake-fetcher mirrors.
- The two login-flow tests (`loginFlowWithLoadingTransition`, `loginFlowFailsWithoutWait`) fail **identically on `main`** with the same test-path fix applied (step 4 `tap --label 'Sign In'` matches nothing, while E2E `TypeTests` pass 11/11). Pre-existing playground-flow issue, unrelated to this PR — recommend a follow-up issue.

### Tests

- `make build` OK; `make test` — 589 tests / 118 suites passed (unit).
- E2E: `BatchTests` 12/14 (see above), `TypeTests` 11/11 on the same simulator.

PR #21, opened 2026-07-02
Surface BGRA stream-video failures as non-zero exit

onevcat:
## Summary

Fixes deferred review item D12: `ios stream-video --format bgra` could exit 0 having streamed nothing.

The BGRA path only wrote `startStreaming` failures to stderr from the `FBFuture` completion callback — the error was never thrown and never affected the exit code. Errors arriving after the 1-second startup window (mid-stream death) were ignored entirely.

## Changes

- **`FirstErrorBox`** (`Sources/iOSSimBackend/Util/VideoCommandSupport.swift`) — thread-safe, set-once error capture, placed next to `CancellationFlag` which it complements. The FB futures complete on a background queue with no continuation to resume; the command loop polls the box instead.
- **`streamBGRA`** (`Sources/iOSSimBackend/Verbs/IOSSimStreamVideoCommand.swift`) — boxes the first error from both the private `startStreaming` future (which per the idb source can resolve with an error at any point, not just during startup) and the operation's `completed` future (the mid-stream termination channel). The command now:
  - fails fast after the existing 1 s startup grace period if the start future already errored;
  - breaks out of the streaming wait loop when an error arrives later;
  - throws the boxed error *before* attempting the graceful stop — `stopStreaming` on a dead stream fails with a secondary error that would mask the original.
- **Help text** — `--format` now notes bgra is experimental and reports no frame count.
- **`Tests/FirstErrorBoxTests.swift`** — pins the box semantics: empty-until-set, read-after-set, first-error-wins, and concurrent sets electing exactly one stable winner.

Deliberately **not** changed (accepted rough edges per review): the fixed 1 s startup sleep, the never-closed stdout `FBFileWriter` (fd stays open by design with `closeOnEndOfFile: false`), and the hardcoded zero-frame `ExecutionResult` for BGRA — the experimental note in the help covers the missing summary.

## Testing

- `make build` + `make test`: 583 tests / 116 suites green (baseline 579/115 + 4 new).
- The streaming happy path and live failure injection need a booted simulator, which is not available in this environment — **no E2E verification was performed**. The error-propagation logic is exercised only through the `FirstErrorBox` unit tests; a manual spot-check of `ios stream-video --format bgra` against a live simulator is recommended before release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Daemon client now retries a command once against the same daemon when the simulator reports the post-boot `transient_booting` readiness gap, matching the long-documented behaviour.
 - Bridge `/swipe` now accepts durations up to 10 s (previously silently clamped to 5 s), covering the full `--duration` range the CLI validates for long-press holds. Bridge `versionCode` bumped to 16.
 - `ios type` builds one HID session for the whole string instead of re-initialising FBSimulatorControl per character.
+- `ios stream-video --format` help now marks `bgra` as experimental (no frame count is reported for that format).
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


onevcat:
## Review

**Verdict: ✅ Approve — no changes needed.** Also performed the live spot-check the PR description asked for.

- **Bug confirmed real**: the old BGRA path wrote `startStreaming` failures to stderr from the future callback and nothing else — no throw, exit 0; mid-stream termination (`completed` future) wasn't observed at all, so a dead stream just spun in the wait loop until a signal.
- **Fix is sound**: `FirstErrorBox` is the right shape for callback-world errors with no continuation to resume; the loop polls it alongside the cancellation flag, and throwing *before* `stopStreaming` correctly avoids masking the original error with the secondary stop failure. Set-once semantics are properly locked and unit-pinned (including the concurrent-election test).

### Live verification (booted iPhone 17 Pro, which the PR author lacked)

- **Happy path**: streamed 4 s → ~12.7 MB of BGRA on stdout, SIGINT → `Stopping BGRA stream... / BGRA stream stopped`, **exit 0**. Graceful path unregressed.
- **Early-closed pipe** (`| head -c 200000`): exit 141 (SIGPIPE) — kernel-level, never a false 0.
- **Simulator shutdown mid-stream**: the process dies with SIGTRAP (exit 133) from inside FBSimulatorControl before any FBFuture error callback fires, so the new `completed`-future path doesn't get a chance to run in this particular scenario. Not a regression and not addressable at the CLI layer — noting it for the record: the box path remains exercised only by unit tests; whatever failure modes FB delivers via futures (attach failures, consumer errors) now exit non-zero, and nothing observed can produce a false success.

### Tests

`make build` OK; `make test` — 583 tests / 116 suites passed.

PR #23, opened 2026-07-02
fix: gate cached HID connection reuse on simulator boot identity

onevcat:
## Problem

`HIDInteractor` keeps a per-UDID cache of `FBSimulatorHID` connections with no liveness check on reuse. A simulator that is shut down and **re-booted under the same UDID** passes `makeSession`'s `state == .booted` re-check, so the daemon hands every verb the cached connection whose mach port belongs to the previous boot.

The existing invalidation path (`DaemonDispatch.platformStaleCleanup` → `clearHIDConnection(for:)`) only fires when a verb error matches `DaemonErrorKind.isStaleSimulatorMessage` ("not found in set" / "is not booted. Current state"). The dead-connection failure matches neither, so it was classified `.other` and the UDID stayed broken until daemon restart.

**Live repro made it worse than reported**: on Xcode 26.5 / iOS 26.2, sending through the dead port doesn't error at all — SimulatorKit never invokes the send completion, so `event.perform` **hangs the daemon mid-request indefinitely** (verified with a wedged daemon; last log line is "Performing HID event...", main thread idle in the run loop). No after-the-fact error classification can catch that.

## Fix

Two layers:

**1. Pre-send boot-identity gate (handles the reboot window, including the hang).**
Each cache entry records the mtime of `<dataDirectory>/var/run/launchd_bootstrap.plist`, which CoreSimulator rewrites on every boot (public API only — `FBSimulator.dataDirectory` + one stat per cache hit). On reuse, a token mismatch — or an unreadable marker, failing closed — discards the entry and builds a fresh connection before anything is sent. Rebuilding costs milliseconds; reusing a dead port hangs forever.

**2. Fail-invalidate + cautious retry-once (handles dead transports that do surface errors).**
Any failed perform now always drops the cached connection, so the next command self-heals. The event is re-performed once, against a freshly built session, **only** when the error message proves nothing reached the currently-booted simulator instance:

- `"Mach port invalid, device disconnected"` — SimulatorKit's `SimDeviceLegacyHIDClient.HIDError` for `MACH_SEND_INVALID_DEST` (taxonomy recovered from the Xcode 26.x SimulatorKit binary: nil port / `MACH_SEND_INVALID_DEST` / other kern returns map to three distinct wordings)
- `"Mach port not connected, device may not be ready yet"` — client never obtained its HID port; nothing was ever sent
- `"has already been disposed"` — `FBSimulatorHID.connect` after disconnect, emitted before any send
- `"(ipc/send) invalid destination port"` — defensive, `mach_error_string` rendering of the same kern return

### Retry-safety reasoning

`FBSimulatorHIDEvent` composites (tap = down + up, etc.) send one mach message per sub-event and short-circuit on the first failure. For the gated errors above, the kernel refused (or the client never had a route for) the failing message, and any earlier sub-events either failed identically or were delivered to the *previous* boot, which no longer exists — so a single re-perform cannot double-apply on any live UI. Everything else — notably SimulatorKit's generic `"Mach return error <code>"` (e.g. a send timeout hitting the middle of a composite whose earlier touches were delivered) — stays invalidate-only, consistent with the recent daemon ambiguous-outcome work that avoids double-applied side-effecting verbs.

## Testing

- 18 new unit tests (`HIDBootIdentityTests`, `HIDPerformRecoveryTests`): token derivation from a temp data-directory fixture, reuse truth table (unknown tokens fail closed), error classification, and recovery orchestration (invalidate-before-retry ordering, no-retry rethrow, retry-error propagation) — decision logic is factored into `HIDBootIdentity` / `HIDPerformRecovery` with injected closures so no FB* types are needed in tests.
- Full suite: **597 tests / 119 suites green** (baseline 579/115).
- Live end-to-end on iPhone 17 Pro (iOS 26.2): `button home` → `simctl shutdown` + `boot` with the daemon surviving → `button home` succeeds in ~1s via "Boot token changed … discarding cached HID connection" → fresh connection. Previously this hung the daemon indefinitely.

## Known limitation

A port that dies *without* a reboot (same boot token, e.g. HID stack breakage inside a still-booted sim) and hangs rather than errors is not covered; that would need a perform watchdog, which risks interacting with legitimately long performs (swipes/holds up to 10s) and is left out of scope here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Cached HID connection is now validated against the simulator's boot instance before reuse, so a simulator shut down and re-booted under the same UDID gets a fresh connection instead of hanging the daemon (or failing every command) on the dead one until restart. Additionally, any failed HID perform drops the cached connection, and failures that provably happened before delivery (dead mach port) are transparently retried once against a rebuilt session.
 - Daemon client no longer tears down a healthy daemon and re-executes the command when the daemon answers with a command-level error (element not found, etc.). Failed commands now surface immediately instead of paying a full daemon respawn, and side-effecting commands are no longer executed twice.
 - Daemon shutdown no longer deletes the socket/pidfile of a successor daemon that took the paths over, which previously chained invisible orphan daemons.
 - Daemon base directory under `/tmp` is now validated on every run (symlinks, foreign owners rejected; loose permissions tightened to 0700) instead of trusting whatever was pre-created there.
--- a/Sources/iOSSimBackend/HID/HIDBootIdentity.swift
+++ b/Sources/iOSSimBackend/HID/HIDBootIdentity.swift
@@ -1,0 +1,40 @@
+// SPDX-License-Identifier: Apache-2.0
+import Foundation
+
+/// Boot-instance identity for the per-UDID HID connection cache.
+///
+/// A cached `FBSimulatorHID` is only usable against the boot it was
+/// created for: its mach port dies with `launchd_sim`. A simulator that
+/// is shut down and re-booted under the same UDID still passes
+/// `makeSession`'s `state == .booted` re-check, and on current
+/// SimulatorKit a send through the dead port never invokes its
+/// completion (observed live on Xcode 26.5 / iOS 26.2), so the failure
+/// cannot even be caught after the fact — reuse has to be gated before
+/// anything is sent.
+///
+/// The token is the modification date of
+/// `<dataDirectory>/var/run/launchd_bootstrap.plist`, which
+/// CoreSimulator rewrites on every boot. Reading it is a single stat —
+/// cheap enough to run on every cache hit — and needs no private API
+/// (`FBSimulator.dataDirectory` is public surface).
+enum HIDBootIdentity {
+
+    /// Whether a cached connection may be reused. Unknown tokens fail
+    /// closed: rebuilding the connection costs milliseconds, while
+    /// reusing a dead one hangs the daemon until restart.
+    static func isReusable(cachedToken: Date?, currentToken: Date?) -> Bool {
+        guard let cachedToken, let currentToken else { return false }
+        return cachedToken == currentToken
+    }
+
+    /// The current boot token for a simulator, or nil when the marker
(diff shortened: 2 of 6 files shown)


onevcat:
## Review

**Verdict: ✅ Approve — no changes needed.** Reproduced both the bug and the fix live.

### Live verification (iPhone 17 Pro, iOS 26.2, Xcode 26.5)

- **Boot-token premise**: statted `<dataDirectory>/var/run/launchd_bootstrap.plist` across a `simctl shutdown` + `boot` cycle — mtime changed (21:28:22 → 21:34:32). The token really does track boot identity.
- **Bug reproduced on `main`**: `ios button home` (daemon caches the HID connection) → reboot the simulator with the daemon surviving → second `button home` **hangs indefinitely** (killed after a 15 s bound; the wedged daemon had to be SIGKILLed). Exactly the reported failure mode — no error ever surfaces for after-the-fact classification to catch.
- **Fix verified on this branch**: same sequence → second press succeeds in **0.83 s**, daemon log shows `Boot token changed for simulator …; discarding cached HID connection`, daemon stays healthy across further presses.

### Code review

- **Layer 1 (boot-token gate)** is the right shape: check-before-send is the only thing that can beat a hang, unknown tokens fail closed (one ~ms rebuild vs. an unrecoverable wedge), and the one TOCTOU wrinkle (token read before `connectToHID`) errs in the safe direction — a mid-window reboot at worst discards a live connection next time.
- **Layer 2 (fail-invalidate + retry-once)**: the invariant "a failed perform never leaves its connection cached" holds on both the primary and the retry path; `recover` guarantees invalidate-before-rebuild; the retry calls `performHIDEventOnce` so there's no recursion. The retry gate's double-apply reasoning (only errors proving non-delivery to the current boot; generic `Mach return error <code>` stays invalidate-only) is consistent with the daemon's ambiguous-outcome policy for side-effecting verbs.
- String-matching SimulatorKit error wordings is inherently version-brittle, but the taxonomy is documented, the default is the conservative branch, and a wording drift degrades to today's behavior (invalidate-only) rather than anything unsafe.
- The stated limitation (same-boot port death that hangs rather than errors) is fair to leave out — a perform watchdog would fight legitimately long composites.

### Tests

`make build` OK; `make test` — 597 tests / 119 suites passed. The 18 new unit tests cover the token truth table, marker fixtures, classification, and orchestration ordering.

PR #24, opened 2026-07-02
Swipe coordinate pairs, whitespace-tolerant selector matching, and a full-screen tap advisory

joonyeonglim:
## Motivation

While using sim-use to drive **Flutter** apps on the iOS Simulator I hit three
small friction points. Each fix is additive and back-compatible; they're split
into two commits (swipe ergonomics; selector robustness).

### 1. Ergonomic `swipe` coordinates
`swipe` required four separate options (`--start-x --start-y --end-x --end-y`)
while `tap` takes `-x/-y` and `describe-ui` takes `--point x,y`. This adds
`--from x,y --to x,y` and positional `x,y x,y` across the top-level, iOS,
Android, and iOS batch surfaces. The existing four flags keep working; exactly
one form must be fully specified.

### 2. Whitespace-tolerant selector matching
On canvas-rendered UIs the accessibility tree often exposes **multi-line**
`AXLabel`s. The compact `describe-ui` outline renders them space-joined, so the
string an agent copies back into `--label` never equals the newline-bearing
label under exact comparison. `--label`/`--value` exact matching and
`--label-contains` now fall back to a whitespace-collapsed comparison **only
when the exact pass finds nothing**, so existing exact matches are unaffected.

### 3. Full-screen tap advisory
Frameworks like Flutter wrap large regions in a single accessibility element
whose frame spans the whole screen. A `--label`/`--value`/`--label-contains`
selector that resolves to such a wrapper taps the **screen centre**, silently
missing the intended control. When a match covers ≥90% of the screen,
`resolveCenterPoint` now emits a non-fatal `[i]` advisory to **stderr** (so
`--json` stdout is never corrupted) recommending a positional `@N`/`#N` alias or
explicit coordinates. The tap still fires — behaviour is unchanged, only a
warning is added.

## Back-compatibility
All three changes are additive. No existing flag, output, or match result
changes: the swipe flags still work, exact label matches are untouched (the
whitespace fallback only runs on zero exact matches), and the advisory is a
stderr-only diagnostic that does not alter the tap.

## Tests
New: `Tests/WhitespaceLabelMatchTests.swift`, `Tests/FullScreenTapAdvisoryTests.swift`
(pure decision logic + resolver behaviour), plus extended swipe tests. Full
suite green locally: **592 tests / 117 suites pass** (`make build && make test`).

Commits are DCO signed off.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `swipe` now accepts `--from x,y --to x,y` and positional `x,y x,y` coordinates on top-level, iOS, Android, and iOS batch surfaces while keeping the existing four coordinate flags.
+
 ### Changed
 
 - Daemon client now retries a command once against the same daemon when the simulator reports the post-boot `transient_booting` readiness gap, matching the long-documented behaviour.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `swipe` now accepts `--from x,y --to x,y` and positional `x,y x,y` coordinates on top-level, iOS, Android, and iOS batch surfaces while keeping the existing four coordinate flags.
+- `tap`/`long-press` now print a non-fatal `[i]` advisory to stderr when a `--label`/`--value`/`--label-contains` selector resolves to an element covering ≥90% of the screen — a common footgun on canvas-rendered UIs (e.g. Flutter) where a full-screen wrapper element makes the tap land on the screen centre instead of the intended control. The tap still fires; the advisory recommends a positional `@N`/`#N` alias or explicit coordinates.
 
 ### Changed
 
+- `--label`/`--value` exact matching and `--label-contains` now fall back to whitespace-collapsed comparison when the exact pass finds nothing, so a multi-line `AXLabel` (which the compact `describe-ui` outline renders space-joined) matches the space-joined string an agent copies back. Existing exact matches are unaffected — the fallback only runs when the exact pass matched zero elements.
 - Daemon client now retries a command once against the same daemon when the simulator reports the post-boot `transient_booting` readiness gap, matching the long-documented behaviour.
 - Bridge `/swipe` now accepts durations up to 10 s (previously silently clamped to 5 s), covering the full `--duration` range the CLI validates for long-press holds. Bridge `versionCode` bumped to 16.
 - `ios type` builds one HID session for the whole string instead of re-initialising FBSimulatorControl per character.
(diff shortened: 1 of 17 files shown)


onevcat:
Hi @joonyeonglim — thank you for this PR! This is a thoughtful set of fixes, and all three target real pain points we've hit ourselves. I went through the description carefully and everything checks out: the flag-shape inconsistency, the space-joined outline round-trip, and the full-screen-wrapper footgun are all real, and the test suite passes as stated. The overall direction is very welcome. 👍

A few structural requests before merging:

**1. Could you split this into three separate PRs?**
The three changes are independent dimensions (swipe ergonomics / selector matching / tap advisory) and we'd like to discuss them at their own pace. The first two look close to mergeable already; the advisory needs a design round (see below), and splitting keeps that from blocking the rest.

**2. Swipe coordinate pairs** — nice ergonomics, and the "exactly one complete form" validation reads well.
Optional cleanup: the coordinate flag declarations are now copy-pasted across `Swipe` / `IOSSimSwipeCommand` / `AndroidSwipeCommand` — a shared `@OptionGroup` (like the existing `UDIDOptions` / `JSONOutputOptions`) would keep them in lockstep.

**3. Whitespace-tolerant matching** — design LGTM; the zero-exact-match gate and the empty-needle guard on `--label-contains` are exactly right. Happy to take this as its own PR essentially as-is.

**4. Full-screen tap advisory — please rework this one on top of current `main`.**
Two reasons:

- The branch is based a fair number of commits behind `main` and currently conflicts with it (`IOSSimSwipeCommand.swift`, `CHANGELOG.md`), so a rebase is needed anyway.
- More importantly, the advisory as written never reaches the user in the default path: `tap`/`long-press` execute inside the auto-spawned per-UDID daemon, whose stderr is redirected to the per-UDID logfile (`DaemonClient.swift`, `task.standardError = logHandle`), and the wire protocol only carries the structured response — so the `[i]` line lands in `/tmp/sim-use-<uid>/<udid>.log` instead of the caller's terminal. The unit tests pass because they exercise the pure function in-process.

We already have the right transport for exactly this kind of signal: the process-liveness advisory (`ProcessAdvisory` / `ProcessAdvisoryRenderer`). It travels as a structured field in the daemon success envelope, gets rendered client-side above the command output, and appears in `--json` under its own key — the Android crash-dialog banner rides the same idea into the describe-ui outline. We'd like the full-screen advisory to take that path so it reaches both the text surface and `--json` consumers.
While reworking it, it's also worth revisiting the screen-area denominator: `roots.first?.frame` isn't guaranteed to be the display bounds when multiple AX roots are present (keyboard/alert windows), which can misfire the ≥90% heuristic in both directions.

Thanks again — happy to review the split PRs quickly once they're up. 🙏

joonyeonglim:
Superseded by the split PRs requested in review:

- #27: Add ergonomic swipe coordinate pairs
- #28: Add whitespace-tolerant selector matching
- #29: Surface full-screen tap advisory through the success envelope

I’m closing this combined PR in favor of those focused PRs. Local verification and live spot checks are listed in each split PR body. Current GitHub status: DCO is passing on all three split PRs, but the `Tests` workflow is still waiting for fork-workflow approval (`action_required`), so GitHub CI has not run yet.

PR #28, opened 2026-07-03
Add whitespace-tolerant selector matching

joonyeonglim:
## Summary
Split from #24. This makes selector matching tolerate whitespace differences only after exact matching fails.

- `--label` / `--value`: exact match first, then whitespace-collapsed comparison if no exact match exists
- `--label-contains`: same fallback behavior for contains matching
- Empty collapsed contains needles do not match everything

## Tests
- `swift test --filter WhitespaceLabelMatchTests`

Signed-off-by: joonyeonglim <lyt970120@gmail.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - JSON output no longer emits the legacy `udid` key (dual-emitted since the `deviceId` transition); `deviceId` is the canonical key in `devices --json`, `daemon stop/status --json`, and Viewer API responses. Inputs (daemon wire decode, Viewer API requests) still accept `udid` as a deprecated alias, to be removed in a future release.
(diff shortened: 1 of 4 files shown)


onevcat:
Thank you for this one too, @joonyeonglim! 🙏

This merged into `main` through #31 (GitHub marked it merged automatically because #31 was built directly on top of your commit `648e7b9`, authorship preserved). The design — the exact-first gate and the empty-needle guard — shipped exactly as you wrote it; on top of it we generalized the mechanism based on review:

- the collapse + exact-first/collapsed-fallback policy moved into a shared `SelectorTextMatcher` in SimUseCore, now also used by `AndroidSelectorResolver` (`--label`/`--label-contains`/`--value`/`--value-contains`) so the copy-from-outline workflow works identically on Android
- both outline renderers delegate their display collapse to the same helper, making the round-trip hold by construction
- `--wait-timeout` polling now retries transient `multipleMatches` that the fallback can surface mid-transition
- the round-trip claim is scoped to labels the outline renders untruncated/unescaped, with `@N` / `describe-ui --json` recommended for long labels

Your tests run unchanged in CI. Looking forward to the advisory PR from the #24 split!

PR #32, opened 2026-07-04
Add preflight device option regression test

BoWuGit:
## Summary

Adds a regression test for the bundled `sim-use` skill preflight script so device discovery stays device-option free.

The released 0.9.0 bundled skill can fail when `preflight.py --device <UDID>` appends `--device` to `sim-use devices --json`, because `devices` is a global listing command and does not accept device-scoped options. Current `main` already fixes that behavior; this test locks it in before the next release.

## What Changed

- Adds `PreflightScriptTests.deviceListingDoesNotReceiveDeviceScopedOptions`.
- Uses a fake `sim-use` binary that fails if `devices --json` receives `--device`.
- Verifies the later `ui --json` command still receives the selected device.

## Validation

- `./scripts/build.sh dev`
- `python3 -m py_compile skills/sim-use/scripts/preflight.py`
- Manual fake-binary preflight run confirmed:
  - `devices --json`
  - `ui --json --device target-device`
- `swift test --filter PreflightScriptTests`
- `make build`
- `make test` (672 tests in 132 suites passed)

### DIFF
--- a/Tests/PreflightScriptTests.swift
+++ b/Tests/PreflightScriptTests.swift
@@ -1,0 +1,63 @@
+// SPDX-License-Identifier: Apache-2.0
+import Foundation
+import Testing
+
+@Suite("Preflight Script Tests")
+struct PreflightScriptTests {
+    @Test("device listing does not receive device-scoped options")
+    func deviceListingDoesNotReceiveDeviceScopedOptions() async throws {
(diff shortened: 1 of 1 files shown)


onevcat (approved):
LGTM — clean regression test for the preflight device-option bug. The only CI failure (AdbRunnerTests flaky stderr assertion) is unrelated and fixed in #33.

PR #39, opened 2026-07-08
fix: drain simctl stdout/stderr to avoid waitUntilExit deadlock

### LINKED ISSUE #37: sim-use devices hangs forever: stdout pipe not drained before waitUntilExit
## Summary

`sim-use devices` hangs indefinitely with no output on a machine that has a lot
of simulators installed. `runSimctl` only reads simctl's stdout after
`waitUntilExit()` returns, so once the JSON output is larger than the ~64 KB pipe
buffer, simctl blocks on write, never exits, and we wait on it forever.

The Android side (`Adb.run`) already handles this case correctly; the simctl path
just doesn't.

## Environment

- sim-use 0.9.0 (Homebrew, lycorp-jp/tap)
- macOS 26.3.1, Apple Silicon
- 124 simulators across 9 runtimes (iOS 15.4-26.4); `xcrun simctl list devices -j`
  is 70,719 bytes here

## Repro

Any machine where `simctl list devices -j` prints more than ~64 KB:

```bash
sim-use devices    # never returns, no stdout/stderr
```

`--version` and `init` are fine since they don't spawn a child.

## Cause

`Sources/iOSSimBackend/Sim/SimctlDeviceLister.swift`, `runSimctl`:

```swift
process.standardOutput = stdout   // Pipe()
process.standardError = stderr    // Pipe()
try process.run()
process.waitUntilExit()                                    // waits before reading
...
return stdout.fileHandleForReading.readDataToEndOfFile()   // drained only after exit
```

Sampling both processes confirms it: `sim-use` is parked in `waitUntilExit`, and
the `simctl` child is blocked in `__write_nocancel` on fd 1, which lsof reports as
a 64 KB pipe:

```
sim-use:  ... runSimctl -> -[NSConcreteTask waitUntilExit] -> mach_msg
simctl:   ... __sfvwrite -> __write_nocancel        (blocked writing stdout)
lsof:     simctl ... 1 PIPE 0x... 65536
```

Redirecting the same command to a file instead of a pipe works fine
(`xcrun simctl list devices -j > out.json`, 70 KB), which is why this only bites
on machines with enough simulators to cross 64 KB.

## Fix

`Adb.run` in `Sources/AndroidBackend/Adb/Adb.swift` already solves exactly this:
it drains both pipes with `readabilityHandler` while the child runs, and even
calls out "the classic 'Process + Pipe deadlock' trap" in a comment. `runSimctl`
should do the same (and probably grow a timeout while it's at it). Might be worth
pulling the drain-and-wait logic into one shared helper so the two paths can't
drift apart again.

The same wait-before-drain pattern also lives in
`Sources/SimUseCore/DeviceResolver.swift` (`simctlBootedListProvider`, running
`simctl list devices booted -j`). That one is booted-only so its output is small
and unlikely to cross 64 KB in practice, but it has the same shape.

## Note

`sim-use ui` also hangs on this machine, but I did not capture its stack, so it
could be a different problem.

JustHm:
## Summary

`sim-use devices` hangs indefinitely on machines with enough simulators that
`xcrun simctl list devices -j` exceeds the ~64 KB kernel pipe buffer.
`SimctlDeviceLister.runSimctl` reads the pipes only after `waitUntilExit()`, so
once simctl's output overflows the buffer the child blocks on `write(2)` and
never exits — the parent waits on it forever.

This drains stdout/stderr concurrently while the child runs (the same
`readabilityHandler` drain `Adb.run` already uses on the Android path), so the
buffer never fills and simctl always completes, regardless of output size.

Fixes #37

## Changes

- `runSimctl` drains both pipes via `readabilityHandler` while the child runs,
  collecting any remainder after exit. `waitUntilExit()` is kept.
- `runSimctl` made `internal` + injectable `executablePath` so the runner tests
  can drive the drain against `/bin/sh` (no live `xcrun simctl` needed).
- Two regression tests mirroring `AdbRunnerTests` (200 KB stdout / stderr):
  without the drain they deadlock; with it they pass in ~0.07 s.

## Scope / notes

- **Timeout not introduced.** `Adb.run` pairs its drain with a
  `terminationHandler` + semaphore timeout; I kept `waitUntilExit()` here to stay
  surgical, since `runSimctl` has no timeout semantics today. Happy to add
  semaphore+timeout parity if you'd prefer — note the regression tests would then
  fail red instead of hanging until the CI job timeout.
- **Sibling site left alone.** `DeviceResolver.simctlBootedListProvider` has the
  same wait-before-drain shape but runs `simctl list devices booted -j`
  (booted-only, small output), so it's very unlikely to cross 64 KB in practice.
  Left out to keep this PR minimal and focused — happy to follow up separately.

## Testing

- `make test` — full unit suite green (692 tests).
- Real repro: on a machine with 117 simulators (~66 KB `simctl` output), the
  released 0.9.0 `sim-use devices --all` hangs indefinitely; the patched build
  returns in ~0.9 s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `sim-use devices` no longer hangs forever on machines with enough simulators installed to push `simctl list devices -j` past the ~64 KB kernel pipe buffer. `SimctlDeviceLister.runSimctl` now drains stdout/stderr concurrently with the child (the same `readabilityHandler` drain `Adb.run` uses on the Android path) instead of reading only after `waitUntilExit()`, which deadlocked once the child blocked on `write(2)`.
 - The root `sim-use --help` abstract no longer claims the tool is iOS-Simulator-only; it now mentions Android emulators/devices as well.
 - `android swipe` invoked directly now enforces the same coordinate rules as the other surfaces: negative coordinates and identical start/end points are rejected at validate time instead of being forwarded to the bridge.
 - Swipe coordinates are validated as finite and ≤ 100000 on all surfaces, so values like `inf`, `nan`, or `1e19` fail with a clean validation error instead of trapping the daemon in the Double→Int conversion.
--- a/Sources/iOSSimBackend/Sim/SimctlDeviceLister.swift
+++ b/Sources/iOSSimBackend/Sim/SimctlDeviceLister.swift
@@ -30,15 +30,34 @@ public enum SimctlDeviceLister {
 
     // MARK: - Internals
 
-    private static func runSimctl(args: [String]) throws -> Data {
+    // `internal` + injectable `executablePath` so tests can drive the drain
+    // against `/bin/sh`; production uses the default `xcrun`.
+    static func runSimctl(executablePath: String = "/usr/bin/xcrun", args: [String]) throws -> Data {
         let process = Process()
-        process.executableURL = URL(fileURLWithPath: "/usr/bin/xcrun")
+        process.executableURL = URL(fileURLWithPath: executablePath)
         process.arguments = args
 
         let stdout = Pipe()
         let stderr = Pipe()
         process.standardOutput = stdout
         process.standardError = stderr
+
+        // Drain both pipes while the child runs. Once `simctl list devices -j`
+        // outgrows the ~64 KB pipe buffer the child blocks on `write(2)` and
+        // never exits, hanging `waitUntilExit()`. Same drain as `Adb.run`.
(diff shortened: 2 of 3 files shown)


onevcat (approved):
Great catch and a textbook diagnosis — the sample/lsof evidence made this an easy review. The fix mirrors `Adb.run`'s drain exactly and the `/bin/sh`-driven regression tests are much appreciated. Thanks for the contribution!

JustHm:
> Great catch and a textbook diagnosis — the sample/lsof evidence made this an easy review. The fix mirrors `Adb.run`'s drain exactly and the `/bin/sh`-driven regression tests are much appreciated. Thanks for the contribution!

Happy to contribute to this project! Thank you for building and maintaining such a great tool.😆

PR #50, opened 2026-07-16
fix: render viewer snapshots for rotated iOS outlines

thickfive:
## Summary

- Allow the Viewer API to parse rotated iOS outline headers such as `App: SampleApp  874x402  (landscape-right)`.
- Keeps `screen` populated so the viewer canvas renders instead of showing `No snapshot yet`.
- Leaves the original outline text unchanged, so the orientation suffix remains visible in the Viewer outline.

## Changes

- Updated `Sources/SimUse/Viewer/ViewerAPIHandlers.swift`
  - Adjusted `parseAppLine` to accept an optional orientation suffix after the `WxH` screen size.
  - Kept the parsed `appLabel` limited to the app name instead of including the suffix.
- Updated `Tests/ViewerAPIHandlersTests.swift`
  - Added a regression test covering an outline header like `App: SampleApp  874x402  (landscape-right)`.
  - Verified the snapshot response still includes `screen.width`, `screen.height`, and a plain `screen.appLabel`.
- Updated `CHANGELOG.md`
  - Added an Unreleased fix note for rotated iOS Simulator viewer snapshots.

## Testing

- `swift test --filter ViewerAPIHandlersTests`
- `make build`
- Manual viewer check with a rotated iOS Simulator window


onevcat:
Thank you for the fix, @thickfive! 🎉

The regex change is exactly right — verified it against a set of edge cases (labels containing parentheses, labels with `NxM`-shaped digits, the empty-label `App:   WxH` form, and Android headers which never carry the suffix) and they all parse correctly, with the `$` anchor keeping the lazy label capture honest. The regression test through the full snapshot handler path is much appreciated too.

I've resolved the CHANGELOG conflict with `main` (both sides added entries under `Unreleased → Fixed`; kept both) and pushed the merge commit to your branch. Merging once CI is green. Thanks again for the contribution! 🙏

Issue #67, opened 2026-07-24
HID transport selection races dtuhidd's boot-time attach; the cached connection pins the loser for the whole boot


onevcat:
## Context

Since the idb bump (`migrate/idb-bump`), the HID transport is resolved by upstream's auto-selection at `FBSimulatorHID` construction: DTUHID when a `dtuhidd` lives in the simulator's `launchd_sim` subtree, legacy Indigo otherwise. `HIDInteractor` then caches the connection under the boot-identity token, so the selection holds for the rest of that boot.

## The race

`dtuhidd` attaches **0–3 s after boot** on a simulator booted while Device Hub is open (measured 2026-07-24, Xcode 27 Beta 4 / CoreSimulator 1169.1 — see `docs/ai/xxxx-xcode27-support/README.md`, 07-24 addenda). A HID verb that lands inside that window:

1. probes the process tree before `dtuhidd` appears → auto-selection picks **Indigo**,
2. the connection is cached under the current boot token,
3. the wrong transport is now **pinned for the entire boot** — no dead-transport error ever fires (Indigo sends "succeed"), so `HIDPerformRecovery` never rebuilds it.

In the reproducible poisoned state this means silently dead keyboard input (`type` exits 0, delivers nothing); in the fully suppressed state the 2026-07-22 matrix documented, touch dies the same way.

Likelihood is low for interactive use (nobody types within 3 s of boot) but real for agents and scripts that `simctl boot && sim-use type` back-to-back.

## Prior art

AXe's equivalent migration ([cameroncooke/AXe#61](https://github.com/cameroncooke/AXe/pull/61)) gates DTU sends on `launchd_sim` uptime ≥ 10 s (`dtuhidMinimumBootUptime`), addressing the sibling race (DTU selected before `dtuhidd` is ready to serve).

## Proposed fix

Don't trust — or at least don't **cache** — the transport selection while `launchd_sim` uptime is inside the boot-attach window (~15 s, the same criterion the retired #60 guard used). Concretely, in `HIDInteractor.getOrCreateHIDConnection`:

- if `launchd_sim` uptime < 15 s (via the existing `LaunchdSimLocator` probe), build the connection but **skip the cache insert**, so the next command re-derives the selection after the window closes; or
- AXe-style: sleep until the minimum uptime before constructing the transport (simpler, costs latency on every fresh boot).

The skip-the-cache variant costs nothing on the steady-state path and converges within one command of the window closing.

## Diagnostics prerequisite

[...]

Issue #55, opened 2026-07-22
iOS daemon holds a stale HID connection across simulator reboot — tap reports success but is not delivered


onevcat:
## Summary

After a `xcrun simctl shutdown <UDID> && xcrun simctl boot <UDID>` cycle, the per-UDID daemon keeps serving HID verbs through the connection it established against the **previous** boot. `tap` returns `✓ Tap at (x, y) completed successfully` but nothing is delivered to the simulator. `sim-use daemon stop --device <UDID>` (forcing a fresh daemon and a fresh HID connection) immediately restores correct behavior.

`HIDPerformRecovery` / `HIDBootIdentity` apparently do not cover this scenario — no recovery is attempted and no error is surfaced.

## Reproduction

Observed twice on 2026-07-22 during the Xcode 27 Beta 4 investigation:

1. Boot a simulator, run `sim-use tap ...` — works, daemon spawns for the UDID.
2. `xcrun simctl shutdown <UDID> && xcrun simctl boot <UDID>` (daemon keeps running).
3. `sim-use tap --x 344 --y 816 --device <UDID>` → exit 0, "completed successfully", **zero UI effect** (verified via `describe-ui` before/after — AX is served fine, so the outline looks alive while touches go nowhere).
4. `sim-use daemon stop --device <UDID>`, repeat the tap → UI responds.

Environment at the time: Xcode 27.0 Beta 4 / CoreSimulator 1169.1 / iOS 27.0 simulator, but the mechanism (daemon-cached HID client vs. a new boot session) is version-independent in principle. Should be re-confirmed on a stable Xcode 26.x setup first.

## Impact

Agents driving a simulator that gets rebooted mid-session silently lose all touch input while every command keeps reporting success — the worst failure mode for an agent-facing tool. Reboots are common in practice (e.g. the dtuhidd keyboard-suppression bypass recommends exactly a shutdown/boot cycle).

## Fix direction

- Detect the boot-identity change (boot UUID / launchd_sim PID) per HID-verb dispatch, or subscribe to boot-state change, and invalidate + reconnect the cached HID client.
- At minimum: fail loudly instead of sending into a dead connection.

Candidate code: `Sources/iOSSimBackend/HID/HIDBootIdentity.swift`, `Sources/iOSSimBackend/HID/HIDPerformRecovery.swift`, daemon session cache in `Sources/SimUseCore/Daemon/`.

PR #68, opened 2026-07-24
feat: migrate to current upstream idb — Device Hub works via DTUHID transport auto-selection

onevcat:
## Summary

Executes step 2 of the Xcode 27 plan (`docs/ai/xxxx-xcode27-support/README.md`): the pinned idb ref moves from `76639e4d` (2025-05-29) to upstream main @ `1f6943f8` (2026-07-23). This is a migration, not a version bump — upstream Swiftified the HID layer, moved to XcodeGen-generated **static** frameworks, and rewrote the video layer — and it buys the headline feature:

**Simulators booted while Device Hub is open now just work.** The HID transport is selected automatically per boot (DTUHID through dtuhidd's CoreDevice service when the legacy HID is suppressed, classic Indigo otherwise), so the state that v0.11.0's guard could only refuse now delivers `tap`/`type`/`swipe` normally.

## What changed

- **Build** (`scripts/build.sh`): pins `1f6943f8`, generates the Xcode project with XcodeGen (new build dep: `brew install xcodegen`), packages static XCFrameworks with `-allow-internal-distribution`, stages the idb `PrivateHeaders` modules for consumers, and drops FBDeviceControl (never imported). No library evolution → `build_products/` is toolchain-locked; re-run `build.sh dev` after switching Xcode.
- **All four `patches/idb/` patches retired** — superseded upstream or targeting deleted code.
- **`Package.swift`**: private-module maps via `-Xcc -fmodule-map-file`; per-archive `-force_load` for the three ObjC-bearing frameworks + weak-linked CoreSimulator/APT tbds. The entire SwiftBuild-rpath machinery (slice rpaths, `stage-fb-frameworks.sh`, release rpath stripping) is deleted — static linking dissolves the dlopen problem it existed for. Deliberately NOT `-ObjC`: blanket force-loading makes Xcode 26.5-built binaries trip a Swift task-allocator abort (bisected to CompanionUtilities), while no force-load at all drops category-only members (`+[FBControlCoreLoggerFactory osLoggerWithLevel:]` unrecognized selector — both reproduced, see the linking addendum in the work record).
- **API migration**: FBFuture bridging deleted; HID events are Swift enums sent via async `hid.send(event:logger:)`; connections construct `FBSimulatorHID(for:transport:)` directly (upstream's `connectToHID()` cache has no boot-identity gate and would resurrect #55); accessibility consumes the typed `FBAccessibilityElement` API through a legacy-shape bridge (serializer shapes unchanged → outline/recovery/calibration pipelines untouched); video uses the async `createStream(configuration:to:)` lifecycle; multi-touch rides upstream's native `.twoFingerTouch`, assembled as one composite so the DTUHID transport drains once per gesture.
- **Guard retirement**: `DeviceHubHIDSuppression` + `SIM_USE_SKIP_DTUHIDD_CHECK` removed; `SIM_USE_HID_TRANSPORT=indigo|dtuhid` added as a debug override, and connection creation logs the dtuhidd-presence signal the auto-selection keys on. The facebook/idb#941 mitigation (explicit `xcodeFrameworks` preload) is kept.
- **Release path**: payload shrinks to the binary + two resource bundles (no `Frameworks/`, no framework codesigning); formula/staging/dev-install updated accordingly.
- **CI**: caches `PrivateHeaders` alongside the XCFrameworks, keys on the Xcode *build id* (betas share a marketing version but ship incompatible toolchain-locked swiftmodules), installs XcodeGen on cache miss.

## Verification

Both toolchains, full regression: **Xcode 26.5 and 27 Beta 4 each run 1116 unit tests + 21/21 E2E suites, zero warnings, zero failures.**

Live 4-case matrix (deterministic judges: Playground tap counter, Settings navigation, Safari URL field):

| Case | Transport picked | tap | type |
|---|---|---|---|
| Xcode 27 B4 + Device Hub (poisoned boot) | auto → DTUHID | ✅ | ✅ |
| Xcode 27 B4 + classic Simulator.app | auto → Indigo | ✅ | ✅ |
| Xcode 26.5 + classic Simulator.app | auto → Indigo | ✅ | ✅ |
| Xcode 26.5 + Device Hub (poisoned boot) | auto → DTUHID | ✅ | ✅ |

Plus: forced-transport A/B in the poisoned state (forced `dtuhid` delivers; forced `indigo` keyboard fail-louds with upstream's diagnostic), manual pinch-out recognized on the gesture playground (multi-touch has no scripted iOS E2E suite), and the standalone-harness re-validation recorded in the work-record addenda.

## Follow-ups

- #67 — HID transport selection races dtuhidd's boot-time attach (~0–3 s window); the cached connection pins the loser for the whole boot. Includes the diagnostics prerequisite (HID info-lines currently reach no sink).
- Watch items from the AXe comparison (cameroncooke/AXe#61) that we could **not** reproduce on our stack — transient empty AX roots, `.traits`-triggered empty hierarchies — are recorded in the work record with their reference fixes.

Work record with full investigation trail: `docs/ai/xxxx-xcode27-support/README.md`.


onevcat:
**External review finding addressed in 5b0348c** — the deadline-never-fires regression is real and is now fixed.

Verified at code level before fixing:
- `HIDSendDeadline` raced via `withThrowingTaskGroup`, which awaits **all** children before returning; `cancelAll()` is only a cooperative mark.
- Both migrated send paths suspend on bare continuations with no cancellation handler (`FBSimulatorIndigoHIDClient.send` — resumed only by the mach completion; `FBSimulatorDTUHIDTransport.send` — resumed only by `xpc_connection_send_barrier`). A dead port means the child never ends and the timeout error could never propagate.
- The pre-migration bridge did not have this hole: `FutureBridge` forwarded task cancellation to `FBFuture.cancel()`, which resolved the wrapper future even while the underlying mach send dangled.

Fix: unstructured first-resume-wins race — the loser is cancelled and then **abandoned**, so the caller gets the timeout at the deadline regardless (one suspended task leaks per dead send, the same abandonment semantics the FBFuture bridge had). Outer cancellation resumes promptly with `CancellationError`. The timeout path already disconnects and drops the cached connection via `HIDPerformRecovery` → `clearHIDConnection`.

Regression tests use a cancellation-deaf never-resumed continuation as the dead-port stand-in (the old suite's `Task.sleep` honors cancellation and could not catch this); the new timeout test hangs under the previous implementation and returns at the deadline under the fix. 1118 unit tests green on both Xcode 26.5 and 27 Beta 4.

PR #85, opened 2026-07-29
Fix Retina JPEG scaling for stream-video

### LINKED ISSUE #83: fix: stream-video --scale renders via NSImage.lockFocus at the host's backing scale factor (no-op on Retina)
## Symptom

`stream-video --scale 0.5` (JPEG formats: `mjpeg` / `raw` / `ffmpeg`) is supposed to shrink each frame to half its dimensions to save bandwidth and encode time. On a Retina Mac it effectively does **nothing to the pixel count**: the emitted JPEG has (roughly) the same pixel dimensions as the unscaled frame — you pay the full decode + re-encode cost per frame and get none of the size reduction. On a non-Retina display (or a headless CI host) the same command produces genuinely half-size frames. Same flags, different output, depending on the Mac it runs on.

This was noticed while unit-testing the frame pipeline for #81: scaling a 100×60 test frame at `--scale 0.5` produced a 100×60 JPEG on a Retina host instead of the expected 50×30. The current test deliberately pins only "re-encoded + aspect ratio preserved" because the absolute pixel size is environment-dependent.

## Root cause

`VideoFrameUtilities.scaleJPEGData` (Sources/SimUseVideo/VideoCommandSupport.swift) resizes through AppKit:

```swift
let newImage = NSImage(size: newSize)   // newSize is in POINTS
newImage.lockFocus()                    // offscreen context at the MAIN SCREEN's backing scale
image.draw(in: NSRect(origin: .zero, size: newSize))
newImage.unlockFocus()
```

Two unit systems are being conflated:

1. `NSImage.size` is in **points**, not pixels.
2. `lockFocus()` renders into an offscreen context whose backing scale factor follows the host's main screen — **2x on Retina**. Drawing a "50×30 point" image therefore produces a 100×60 **pixel** bitmap, which `tiffRepresentation` then encodes.

So the point-space math is correct, but the pixel output is `points × hostBackingScale` — on a 2x host, `--scale 0.5` round-trips back to the original pixel count. `reencodeJPEGData` (the quality-only path) has the same AppKit dependency but no size math, so it is only exposed to DPI-metadata edge cases, not the scale bug.

## Affected / not affected

| Surface | Affected? |
|---|---|
| `stream-video --scale` JPEG formats, iOS (`IOSSimStreamVideoCommand.swift:174`) | **Yes** |
| `stream-video --scale` JPEG formats, Android (`AndroidStreamVideoCommand.swift:321`) | **Yes** (shares the same helper) |
| `stream-video --format h264` (Android) | No — scale maps to `screenrecord --size` in pixels |
| `record-video`, both platforms | No — native `FBSimulatorVideoStream scaleFactor` / `screenrecord --size`; the screencap fallback draws into a pixel-exact `CGContext` via `H264StreamRecorder` |
| `screenshot` | No — no scaling applied |

## Proposed fix

Replace the AppKit resize with a pixel-deterministic CoreGraphics/ImageIO pipeline, reusing what the file already has:

1. `makeCGImage(from:)` → source `CGImage` (pixel dimensions, no DPI ambiguity).
2. `computeDimensions(for:scale:)` → target pixel size (already exists, already unit-tested, already enforces even dimensions).
3. Draw into a `CGContext` of exactly those pixel dimensions (same pattern as `H264StreamRecorder.append`).
4. Encode with `CGImageDestination` + `kCGImageDestinationLossyCompressionQuality` (also replaces the NSImage/TIFF round-trip in `reencodeJPEGData`, which is an extra full-image copy per frame today).

Benefits: `--scale` means the same thing on every host, the AppKit dependency (and its main-screen coupling) drops out of `SimUseVideo`, and the per-frame TIFF intermediate goes away.

## Test plan

- Tighten `VideoFrameProcessingTests.processScales` from "re-encoded + aspect preserved" to exact expected pixel dimensions (`100×60 @ 0.5 → 50×30`) — it becomes the regression pin once output is deterministic.
- Add a quality-path assertion that re-encode preserves exact input dimensions.
- Live sanity: `stream-video --format ffmpeg --scale 0.5` on a Retina host, decode a frame, confirm halved pixel dimensions and a visibly smaller per-frame byte size.

Refs: #78, #81 (where the behavior was pinned as-is).

EuanTop:
## Summary

Fixes #83.

This updates JPEG frame scaling to produce the requested output pixel dimensions regardless of the host display backing scale. Previously, the macOS `NSImage.lockFocus()` path could render through the Retina backing scale, so `--scale 0.5` could still emit frames at the original Retina-sized dimensions.

The new path decodes to `CGImage`, computes exact scaled dimensions, draws into an explicitly sized `CGContext`, and encodes the result with `CGImageDestination`.

## Changes

- Replace the AppKit-based JPEG scale/re-encode path with ImageIO/CoreGraphics encoding.
- Add an explicit JPEG encode failure error.
- Add tests that assert scaled JPEG output has exact pixel dimensions.
- Add coverage that quality-only re-encoding preserves dimensions and changes output size.

## Validation

- `make build`
- `swift test --filter VideoFrameProcessingTests`
- `git diff --check`
- Live simulator verification:
  - `--scale 1.0`: `1206x2622`
  - `--scale 0.5`: `602x1310`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- JPEG frame scaling now produces the requested pixel dimensions on Retina displays instead of multiplying the result by the screen backing scale (#83).
 - The `SIM_USE_VERSION` build-time override actually reaches the version stamp now. `VersionPlugin` registered its generator command with an empty environment, so the documented "environment variable first" priority was dead code and every build fell back to `git describe` — release builds happened to be correct only because they build on the tagged commit, while `scripts/dev-install.sh` builds stamped a stale describe string. The override is now resolved in the plugin process and baked into the command's argv, which also makes a changed override invalidate SPM's command cache without requiring a source-file edit.
 
 ## [0.12.0] - 2026-07-29
--- a/Sources/SimUseVideo/VideoCommandSupport.swift
+++ b/Sources/SimUseVideo/VideoCommandSupport.swift
@@ -3,10 +3,7 @@ import Foundation
 import AVFoundation
 import ImageIO
 import os
-#if os(macOS)
-import AppKit
 import SimUseCore
-#endif
 
 /// Stop-path watchdog for the `record-video` command.
 ///
@@ -44,6 +41,7 @@ public enum RecordingFinishWatchdog {
 public enum VideoProcessingError: Error {
     case emptyScreenshot
     case failedToDecodeImage
+    case failedToEncodeImage
     case failedToAllocatePixelBuffer
 }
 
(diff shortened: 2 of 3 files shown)


onevcat:
LGTM! Thank you for this.

PR #101, opened 2026-08-21
feat: bracket GIF exports with opt-in START/END marker cards

onevcat:
Supersedes #100, carrying @kws0210's commits forward unchanged — thanks for the feature and the quick opt-in turnaround! This PR adds the remaining review follow-ups plus CI-hang forensics.

## What

- `record-video --gif-markers`: opt-in START/END marker cards (~1 s each) bracketing GIF exports, on all three surfaces (@kws0210's original work, commits preserved).
- Review follow-ups from #100:
  - `gifMarkers` is a required parameter through `ResolvedRecordingOptions` / `RecordingOutputPlan` / `GIFTranscoder` / `AndroidRecordVideoCommand.record`. The opt-in default lives only on the three `--gif-markers` flag declarations, and an unwired internal caller is now a compile error.
  - Focused parity coverage (`Tests/RecordVideoForwarderTests.swift`): default-off and `--gif-markers` parsing on the top-level, iOS, and Android surfaces; the top-level → iOS forwarder copy; the argument shape of `AndroidRecordVideoCommand.record` the Android forwarder compiles against.
  - `skills/sim-use/SKILL.md`'s evidence-GIF entry now documents the flag (the cheatsheet alone did not cover the primary skill entry).

## CI hang investigation

`Unit tests (macOS)` hit the 60-minute job timeout twice on #100. Established so far:

- Not the runner tooling drift: the last green main run (Aug 6) installed the same xcsift 1.3.1, and the adb orphan process also appears on green runs.
- Not locally reproducible: the full suite passes in ~12 s of test time on a dev machine.
- Zero forensics from either timeout: the raw-log artifact was gated on `failure()`, which is false when the job is cancelled by its timeout.

This PR therefore bounds the test step at 20 minutes and uploads the raw `test.log` on cancellation as well — if the hang reproduces on this run, the partial log will name the wedged test. A rerun of the #100 head is in flight in parallel to confirm determinism; prime suspect is the CoreText path in `makeMarkerCard` (first CoreText use in this codebase, and `CTFontCreateWithName` goes through fontd XPC, which a headless runner can plausibly wedge).

## Testing

- `make build` green; `make test` locally: 1299 passed (the single red is an uncommitted local-workspace test edit unrelated to this branch — CI runs the clean tree).
- Live GIF behavior is unchanged from #100's validated head; this PR only removes default arguments, adds tests, and touches docs/CI.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -139,10 +139,13 @@ jobs:
       # depends on, including the sim-use executable the smoke step
       # runs. make picks xcsift up from PATH and condenses the output
       # to a TOON summary itself; SIM_USE_RAW_LOG additionally saves
-      # the raw swift output, kept as a failure-only artifact for
-      # forensics the condensed view can't serve (e.g. grepping daemon
-      # log lines).
+      # the raw swift output, kept as an artifact on failure or
+      # cancellation for forensics the condensed view can't serve
+      # (e.g. grepping daemon log lines, or finding the last test that
+      # started before a hang). The step timeout bounds a wedged test
+      # run well below the job timeout so the artifact still uploads.
       - name: Run unit tests
+        timeout-minutes: 20
         run: make test
         env:
           SIM_USE_RAW_LOG: ${{ runner.temp }}/test.log
@@ -165,8 +168,11 @@ jobs:
             } >> "${GITHUB_STEP_SUMMARY}"
           fi
 
+      # failure() is false when a step or job times out (that's a
+      # cancellation), and a hang is exactly when the partial raw log
+      # matters most — cover both.
       - name: Upload raw logs on failure
-        if: failure()
+        if: failure() || cancelled()
         uses: actions/upload-artifact@v7
         with:
           name: raw-swift-logs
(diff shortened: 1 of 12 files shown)


onevcat:
Root cause of the CI hang identified and (pending this run) fixed.

**What it was NOT** — each ruled out by direct evidence: CoreText/fontd on the headless runner (the marker-card render test *passes* in the captured log); xcsift (1.3.1 rebuilt from source digests the captured log instantly, same as 1.4.2); toolchain/image drift (macOS 26.5.2, Xcode 26.6 17F113, xcsift 1.3.1 identical across green and hung runs; #102 = main + workflow-only change runs green on the current 20260728 image).

**What it was**: the forensics artifact from this PR's bounded run shows XCTest completing 306/306 and exactly three tests started-but-never-finished — the three that drive a real VideoToolbox H.264 encode session via `makeSyntheticMP4`. `main` has two such tests and weeks of green runs; this feature added the third, and all of them start in the same instant under Swift Testing's default parallelism. Three concurrent VT encode sessions deadlock the encoder on GitHub's virtualized macOS runners (4/4 runs), wedging the whole `swift test` process; the mid-line log cutoff at ~2.2 s was the remaining output stuck in the suspended process's stdio buffer, not a slow test. Never reproducible on real hardware.

**Fix**: `@Suite("GIFTranscoder", .serialized)` — at most one encoder session alive at a time, well under a second of extra local runtime. The 20-minute step bound + on-cancel artifact stay as permanent regression insurance.

Issue #120, opened 2026-08-27
daemon: hold the physical-device DTX session across invocations (amortised setup, live element handles)


onevcat:
## Context

Extracted from the #115 design discussion: #115 is scoped to CLI shape and user-facing behaviour (listing parity, device-kind field, top-level routing, capability model) and closes with those; the daemon integration is this issue. Prerequisite: the `ios-device` verbs' `SimUseExecutableCommand` conformance (#108's PR) and the `Platform.iOSDevice` routing from #115.

Today every `ios-device` invocation opens its own DTX session (`DeviceSession.withClient`): `prepare()` → work → mandatory `finish()`. That costs session setup per call, and — more importantly — element handles die with the session, which is why `@N` is impossible on this surface (#113).

## Proposal

Route physical-iOS verbs through the per-UDID daemon (`SimUseExecutableCommand.simulatorUDIDForDaemon`), and have the daemon hold the DTX session open across requests:

- Session setup (`prepare()`) amortised over the daemon's lifetime.
- **Element handles stay valid between invocations** while the daemon lives → `ui` then `tap @N` becomes natively possible; #113's fingerprint proposal shrinks to a drift-detection layer on top (and its Proposal B — hardening simulator `@N` — is independent and unaffected).

## Hazards (why this is its own issue)

- **Teardown ownership inverts.** `AXAuditClient.finish()` is mandatory — skipping it wedges the device (later sessions connect and answer, but element queries come back empty until something performs the cleanup). With `withClient`'s closure scoping gone, the daemon must own `finish()` on idle timeout and SIGTERM, and the wedge-recovery path after a daemon crash needs live verification before this ships.
- `Daemon.Start`'s `livenessProbe` and `DaemonDispatch.staleSimulatorOutcome` are binary android/simulator forks (`PlatformRouter.looksLikeAndroid`) — both need a third arm (or an explicit nil probe) for physical devices.
- Device unplug/replug while a daemon holds the session: staleness detection has no launchd_sim equivalent here; needs a design (AMDevice detach notification, or fail-and-respawn).

## References

- #115 — CLI-shape parent (PRs A–C); this issue is "PR D" from that plan.
- #113 — `@N` via fingerprint re-resolution; interacts as described above.
[...]

Issue #109, opened 2026-08-26
ios-device: spike — text input via settable attributes


onevcat:
## Context

The audit channel has no keyboard/HID path, so `type` / `paste` are out of scope for physical devices today. But the attribute descriptor the daemon accepts carries a `SettableValue_v1` field (`AXAuditAttribute.isSettable`), which suggests a set-value selector exists on the wire. If setting the `Value` attribute works on editable text elements, a `type`-equivalent (replace/set text) becomes possible on this surface without any HID channel.

## Proposal

Time-boxed spike:

1. Observe Accessibility Inspector's DTX traffic against a development-signed app with a focused text field, or probe plausible selectors on the audit service.
2. If a working selector exists, design `sim-use ios-device set-value --label <X> --value <text>` (resolve + set in one session).
3. If not, record the dead end here so it isn't re-investigated.

## Notes

- Even a replace-only primitive (no per-keystroke events) would unlock login/search flows for agents on physical devices.

onevcat:
Confirmed on-device (iOS 26.6, apiVersion 26, protocolVersion 157): `deviceCapabilities` lists `deviceElement:setValue:attribute:`, so the set-value selector exists on the wire as hypothesized. Focus-change events also report per-attribute `SettableValue_v1` flags (see #105 / #107), which tells us exactly which attributes accept writes on a given element. Remaining work: verify a write against an editable text field and design the verb.

Issue #104, opened 2026-08-26
ios-device: expose scroll actions (AXAction 2006/2007)


onevcat:
## Context

`AXAuditAttribute` already defines `scrollUp` / `scrollDown` (daemon action codes 2006/2007, read back from the Actions section of focus events), but no CLI verb sends them. The hierarchy walk in `ui` is deliberately inert (it never scrolls), so content below the fold is unreachable today: an agent can see the visible tree but has no way to bring offscreen elements into it.

## Proposal

Add `sim-use ios-device scroll --direction up|down [--label <scrollable element>]` — resolve the scrollable target (default: the frontmost scroll view / root) and perform the action inside one connection, same session-scoping as `tap`.

## Notes

- AX scroll actions are page-wise, not pixel-wise; document that granularity.
- Like Activate, the action is fire-and-forget — the follow-up `ui` read is the verification.
- Update README and `skills/sim-use/SKILL.md` accordingly.

Issue #42, opened 2026-07-09
Consolidate the tap-family flag surface into shared OptionGroups to eliminate forwarder copy hazards


onevcat:
## Context

The cross-platform verbs duplicate their flag declarations across surfaces (top-level `Tap` / `LongPress`, `IOSSimTapCommand`), and the top-level forwarders bridge them by constructing the backend command via its empty `init()` and assigning every property by hand — `Tap.executeIOSSim` copies 19 fields today.

A `ParsableArguments` property that is never assigned stays in wrapper-definition state, and the first read traps with ArgumentParser's *"can't read a value from a parsable argument definition"* fatal — the same failure class as #41 (`android tap` crashing on a directly-initialized `MultiTouchOptions()`). For the forwarders the hazard is: add a flag to the shared surface, forget one `sub.field = field` line, and **every existing test stays green** — the parity tests (`flagSurfaceParses`, `pointFlagParsesEverywhere`) only pin that the same argv *parses* on every surface; nothing executes the forwarder copy (that path needs a live device). The fatal then fires at runtime on the first read of the un-copied property. #40 avoided this by manual field-by-field review, which is not a mechanism.

## Proposal

Two layers — the structural consolidation is the point of this issue; the reflection test is its guardrail for whatever manual copying remains.

### 1. Structural: shared `@OptionGroup`s

Collect the loose tap-family flags into shared groups under `SimUseCore/Options/`, following the existing `SwipeCoordinateOptions` / `MultiTouchOptions` / `DeviceOptions` / `JSONOutputOptions` precedent:

- **`TapTargetingOptions`** — `-x`/`-y`/`--point`, the five selector flags (`--id`/`--label`/`--value`/`--label-contains`/`--label-regex`), `--element-type`, `--frame`. Whether the positional alias joins the group (ArgumentParser supports `@Argument` inside `ParsableArguments`) or stays per-command is an implementation decision — help ordering is the main consideration.
- **`TapTimingOptions`** — `--pre-delay`/`--post-delay`/`--wait-timeout`/`--poll-interval`. Caveat: `--duration` differs between `tap` (nil default) and `long-press` (0.8 default, different help text), so it likely stays per-command.

[...]

PR #126, opened 2026-08-27
feat: route physical iOS devices through the top-level verbs (capability model)

### LINKED ISSUE #115: CLI shape: make physical iOS devices a first-class target (listing, device-kind field, top-level + daemon routing)
Problem inventory before we design the CLI shape. This issue only catalogues the current state and the inconsistencies; the actual design is deferred to follow-up discussion.

The theme: `sim-use ios-device` is currently an appendage, not a first-class target. If we commit to physical iOS support, it should behave like every other target (with some capabilities legitimately restricted), rather than living off to the side.

### 1. `sim-use devices` listing coverage

| Target | Listed by `sim-use devices`? | Source |
|---|---|---|
| iOS Simulator | ✅ | `simctl list devices` |
| iOS physical device | ❌ **blind spot** | would need `FBDeviceControl` (what `ios-device devices` already uses) |
| Android emulator | ✅ | `adb devices` |
| Android physical device | ✅ | `adb devices` |

`simctl` only knows simulators, so `sim-use devices` never shows a plugged-in iPhone — even though `sim-use ios-device devices` lists it fine through a different code path. Android, by contrast, shows both emulators and physical devices. So the listing is asymmetric: Android physical is visible, iOS physical is not.

### 2. No device-kind field

`Device` (the `sim-use devices` row / `--json` schema) carries `platform ∈ {ios, android}`, plus free-form `state`, `runtime`, `name`, `deviceId`. There is **no field distinguishing simulator / emulator / physical**:

- Android emulator vs physical device is only inferable from the serial shape (`emulator-5554` prefix), not a structured field.
- iOS simulator vs physical has no field either (and physical isn't listed at all today).

The `PLATFORM` column answers "ios or android" but not "simulator, emulator, or real device", which is exactly the distinction that determines available capabilities.

### 3. Physical iOS is not routed like other targets

| Surface | iOS Simulator | Android | iOS physical |
|---|---|---|---|
| Top-level verbs (`ui`, `tap`, `screenshot`, …) | ✅ routed by UDID shape (`PlatformRouter`) | ✅ | ❌ rejected — must use the `ios-device` namespace |
| Per-UDID daemon (auto-spawn, amortised init) | ✅ | ✅ | ❌ — every `ios-device` call opens its own DTX session |
| Namespace | `sim-use ios <verb>` (iOS-only extras) | `sim-use android <verb>` | `sim-use ios-device <verb>` (the *whole* surface) |

- `PlatformRouter` has only `.iOSSim` / `.android` cases; a physical iOS UDID resolves to neither and top-level verbs reject it (pointing at `ios-device`).
- The `ios-device` subcommands are plain `AsyncParsableCommand`s that call `DeviceSession.withClient` directly — they do not implement `SimUseExecutableCommand`, so they bypass the daemon entirely.

This contradicts sim-use's stated design: "one command surface; the UDID shape decides the backend." A physical iOS device is the only target you address through a wholly separate command tree.

## Problems to solve (design deferred)

1. **Listing parity** — `sim-use devices` should surface physical iOS devices (~~probably opt-in given `FBDeviceControl` discovery cost; see the ~5 s attachment-quiescence wait~~), so discovery is uniform across every target. *(Edit 2026-08-27: the opt-in rationale is retracted — measured discovery is ~0.4 s with a device attached, and #117 cut the no-device path to ~1 s, so default-on is plausible; see the measurement comments below.)*
2. **Device-kind field** — add a structured field distinguishing `simulator` / `emulator` / `physical` (name TBD), orthogonal to `platform`. Lets the table and `--json` say "ios physical" vs "ios simulator" vs "android emulator" explicitly, and lets callers reason about which capabilities apply.
3. **First-class routing** — decide whether physical iOS joins `PlatformRouter` + top-level verb forwarding + the daemon, so `sim-use ui` / `sim-use tap` work against a plugged-in iPhone (with the restricted capability set), instead of requiring `ios-device`. This is the biggest question and where "first-class citizen vs restricted appendage" is actually decided.
4. **Capability model** — with physical iOS routed through the top level, unsupported verbs (coordinate tap, swipe, gesture, ~~screenshot,~~ `--json`, …) need a consistent "capability not available on this target" contract rather than today's hard namespace split. *(Edit 2026-08-27: screenshot moved to the supported column — shipped over CoreDevice in #118. The contract also needs to express "supported by the tool, refused by this device" — CoreDevice screen recording is capability-gated per device.)*

## Constraints / facts that bound the design

- Physical iOS has **no geometry** and **no coordinate input** — coordinate tap / swipe / gesture ~~/ screenshot~~ cannot be supported (channel limitation, confirmed on-device). Any first-class routing must degrade these cleanly. *(Edit 2026-08-27: screenshot **is** supported over a second channel — CoreDevice `xcrun devicectl device capture screenshot`, no dev-signing requirement, shipped as `ios-device screenshot` in #118. The audit-channel limitation stands for coordinate input; see the correction comment below.)*
- Physical-iOS element handles **expire between processes**, so `@N` cannot be a cached-coordinate replay (tracked separately in #113).
- Discovery cost: ~~`FBDeviceControl` attachment discovery pumps the run loop up to ~5 s, so unconditionally adding it to every `sim-use devices` call would regress the common simulator/Android path — hence "opt-in" above.~~ *(Edit 2026-08-27: superseded — ~0.4 s with a device attached, ~1 s no-device after #117's empty-grace bail; see the measurement comments below.)*

## Non-goals for this issue

No implementation, no chosen design — this is the problem catalogue we align on first. Follow-ups will design (a) the device-kind field + listing parity, and (b) the routing/daemon integration, likely as separate changes.

onevcat:
## Summary

PR C of the #115 plan — the final step, and the one where "first-class citizen vs restricted appendage" is actually decided. `Platform.iOSDevice` joins `PlatformRouter`, so a physical UDID resolves to a platform instead of `nil`; all 14 top-level verb switches went non-exhaustive, and the compiler enumerated every per-verb capability decision — plus `app-state`, which routed outside the switch pattern and was pulled into the model during review, for 15 device-scoped decisions in all (3 routed, 12 rejected). Closes #115.

```bash
sim-use ui --device 00008130-…            # routes to the audit channel
sim-use tap '#BackButton' --device 00008130-…
sim-use screenshot --device 00008130-…    # CoreDevice, any screen

sim-use swipe 1,2 3,4 --device 00008130-…
# Error: `swipe` is not supported on physical iOS devices: swipes are coordinate HID
# sequences, and the accessibility audit channel exposes no coordinate input or element geometry.
# Hint: Interact through accessibility actions instead: `sim-use ui` reads the outline, …
```

## The capability model

- **Routed (3):** `describe-ui`/`ui`, `tap` (`#<id>` / `--id` / `--label` / `--label-contains` / `--element-type`), `screenshot`. They forward to new typed executor entry points on the `iOSDeviceBackend` verbs (`performUI` / `performTap` / `performScreenshot`, mirroring `IOSSimTapCommand.performTap` — no hand-built command instances, #42 stays closed).
- **Rejected (11 + tap forms):** a new shared `TargetCapabilityError: HintProviding` (modelled on `HIDKeyCommandHelp.androidUnsupportedMessage`) states *why the channel can't carry the verb* plus the nearest alternative — surfaced as the `hint` in `--json`, and as a `Hint:` line in text mode. `tap` gets a per-form decision table (coordinates, `@N`/`#N` aliases, `--value`, `--label-regex`, `--frame`, `--duration`, `--wait-timeout`, multi-touch reject; `--pre/post-delay` are plain waits and are honoured), pure and pinned by tests. The `#N`-list-cell vs `#<id>`-literal boundary reuses `OutlineAliasResolver.parse`, so the alias grammar cannot drift from the simulator's.
- **Honest result shapes**, not faked parity: `describe-ui` carries `"kind":"physical"` and no `raw`/`entries`/`lists`/`screen` (`screen` goes optional; the outline text with `#id`s — byte-identical to `ios-device ui`, summary line included — is the payload). `tap` carries `"kind":"physical"` with the matched element instead of `x`/`y` (`x`/`y` go optional). Wire compat pinned: old `{x,y}` daemon payloads still decode; sim/Android encodings gain no new keys.

## Boundary decisions

- **`DeviceOptions.resolve(allowPhysical:)`** — the 15 device-scoped top-level verbs opt in; `sim-use ios <verb>` keeps the default and still fast-fails on a physical UDID (`PhysicalIOSDeviceError`, copy now pointing at the routed top-level verbs). No more blanket rejection at resolution time for the routed surface.
- **Daemon exclusion, centrally** — physical targets are filtered out of daemon dispatch in `resolveExecutionResult` (one guard, removed when #120 lands): every physical call runs in-process, so a capability rejection can never auto-spawn an idle per-UDID daemon. Pinned by a pidfile test.
- **ECIDs stay a shape-routing non-goal** (documented): a bare ECID is invisible to shape heuristics; the `ios-device` namespace accepts them.
- **Docs flip with the routing** (per the #115 decision): README + SKILL.md present the top-level verbs as the primary physical surface, with an explicit **verb × target capability matrix** so agents never assume capability parity; `ios-device` is demoted to the physical-only peer of `ios`/`android` (ECIDs + tree-tuning flags). "Experimental" branding stays until the capability set grows (#104 scroll, #109 type).

## Verification

- `make build` / `make test` — 0 warnings, **1382 passed** (+13). New `PhysicalDeviceRoutingTests`: all 11 rejections end-to-end through the binary (message + `Hint:` + exit code, plus the `--json` envelope), the tap form decision table (11 rejected forms, 5 routed forms, `#feed@home` stays a literal id), result-shape and wire-compat pins, routed-vs-namespace summary-line parity, `ios` namespace fast-fail, `PlatformRouter` mapping, and the daemon-exclusion pidfile check.
- Live, simulator regression: `ui` / `tap @1` / `tap --json` against a booted iPhone 17 Pro Max — output byte-identical to before (`✓ Tap at (219.0, 32.0) completed successfully`, `{"data":{"x":219,"y":32},"ok":true}`).
- Live, rejection paths: `swipe` (text + `--json`), `tap @1`, and `ios describe-ui` against a physical UDID — per-verb reasons, hints, exit 1, as quoted above.
- Live, routed success paths — **done** (iPhone One, iOS 26.6, USB, dev-signed Playground foreground):
  - `sim-use ui --device <physical>` — full tree (47 elements / 125 nodes), text byte-identical to `ios-device ui` including the summary line; `--json` `data` keys exactly `{platform:"ios", kind:"physical", outline, entries:[], lists:[], appLabel, appPackage}` — no `screen`, no `raw`, as designed.
  - `sim-use tap --label-contains "Tap Test" --element-type Button --json` → `{"action":"Activate","kind":"physical","label":…,"role":"Button"}` (no `x`/`y`); `sim-use tap '#BackButton'` → `Sent Activate to 'sim-use Playground' [Button] #BackButton`; follow-up `ui` confirmed the navigation both ways.
  - `sim-use screenshot --device <physical>` — valid 1290×2796 PNG, text and `--json` (`{path}`) both green.


onevcat:
## Adversarial review — two rounds, all findings resolved

Ran two independent review rounds against this branch (separate reviewer session, xhigh effort, with parallel routing / wire-compat / regression sub-reviews). Every finding was verified before fixing; both rounds re-verified the previous round's fixes live.

**Round 1** (against `b2be4f4`):
- **P1 — `app-state` bypassed the capability model** (`AppState.swift`): it routed on the legacy `looksLikeAndroid` binarisation outside the `PlatformRouter` switch, so a physical UDID got the generic `PhysicalIOSDeviceError` — whose "only drives iOS Simulators" copy is wrong for a verb that also serves Android — instead of a per-verb rejection. → Fixed in `038d761`: `allowPhysical` + a `.iOSDevice` arm throwing `TargetCapabilityError` (no process-list channel; crash detection is daemon state) with a `sim-use ui` hint; added to the end-to-end rejection table. Round 2 re-verified text/`--json`/`SIM_USE_DEVICE` paths and the Android/`.none` fallbacks live.
- **P2 — `paste` ran the simulator soft-keyboard preflight for physical targets** (`Paste.swift`): the guard only skipped Android, costing ~0.24 s of FBSimulatorControl/AX setup before the rejection. → Fixed in `136b9c8`: preflight keys off the same `PlatformRouter` decision as `execute()`; measured 0.06 s post-fix (re-verified in round 2).
- **P3 — stale Viewer rationale + coverage gaps** → Fixed in `136b9c8`: Viewer comment and `--no-physical-ios` help now state the real product limitation (coordinate taps / frames / streaming); added `SIM_USE_DEVICE`-carrying-physical routing regression and a pre-#115 `describe-ui` payload decode fixture. The suggested fake-binary daemon-lifecycle harness was judged not worth the cost for now — the pidfile assertion covers the observable contract.

**Round 2** (against `038d761`): P1 none. Remaining findings were documentation drift, fixed in `b35253e`:
- **P2** — the skill's preflight section still said "do not route [physical devices] through these top-level commands", contradicting its own physical-devices section post-routing.
- **P3 ×2** — CHANGELOG said "14 top-level capability decisions" (now 15 with `app-state`: 3 routed, 12 rejected), and the #111 entry still justified the Viewer filter with "aren't operable through the top-level verbs", which this branch made false.

Round 2 also independently confirmed: no daemon spawn for physical shapes (modern + legacy), no `SIM_USE_IN_DAEMON` recursion bypass, `sim-use ios …` / `daemon start` still reject physical targets, ECID non-goal consistent, and no Codable compatibility break in the reshaped tap/describe-ui envelopes.

PR #127, opened 2026-08-28
fix: reject incompatible CLI in skill preflight

BoWuGit:
## Summary

- add an explicit minimum CLI compatibility check before device discovery
- report the Homebrew upgrade command when the bundled skill is newer than the installed CLI
- mention physical iOS devices in the device-discovery remediation
- cover compatible and outdated CLI paths in `PreflightScriptTests`

## Motivation

The official `sim-use init` path installs the Skill bundled with the executable and therefore keeps the two versions paired. The mismatch can still occur when a user manually installs the Skill from source or a cross-harness Skill manager updates the repository content independently of the CLI. In that case the v0.14 Skill can document physical iOS routing while a v0.13 CLI silently omits physical devices, and the current preflight misdiagnoses the result as a disconnected device even when `devicectl` can see it.

This check is defense in depth for externally managed Skill installations; it does not replace or change the recommended `sim-use init` flow.

## Verification

- `python3 -m py_compile skills/sim-use/scripts/preflight.py`
- standalone fake-CLI scenarios for v0.14 success and v0.13 early failure
- real physical iPhone preflight with sim-use v0.14.0
- `./scripts/build.sh dev`
- `swift test --filter PreflightScriptTests` — device-list test and all 8 version cases passed
- GitHub Actions: DCO, bridge protocol parity, bridge unit tests, and macOS unit tests all passed

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -3,7 +3,7 @@
 ## Code quality
 - No `Any` types unless absolutely necessary; upgrade dependencies rather than downgrading code to work around type errors.
 - Follow Swift best practices and match the style of surrounding code.
-- When adding or removing commands/options, update the README and `skills/sim-use/SKILL.md`.
+- When adding or removing commands/options, update the README and `skills/sim-use/SKILL.md`. If the Skill starts relying on behavior newer than its current compatibility floor, also bump `MINIMUM_SIM_USE_VERSION` in `skills/sim-use/scripts/preflight.py` and extend `PreflightScriptTests`.
 
 ## Changelog
 `CHANGELOG.md` follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/). Add entries under `## [Unreleased]` as you land changes. Subsection order: `### Added` / `### Changed` / `### Fixed` / `### Removed`. Never modify already-released version sections.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- The bundled skill preflight now rejects an older `sim-use` CLI before device discovery and prints the Homebrew upgrade command, instead of misdiagnosing newly documented device types as disconnected.
+
 ## [0.14.0] - 2026-08-27
 
 ### Added
--- a/Tests/PreflightScriptTests.swift
+++ b/Tests/PreflightScriptTests.swift
@@ -6,70 +6,98 @@ import Testing
 struct PreflightScriptTests {
     @Test("device listing does not receive device-scoped options")
     func deviceListingDoesNotReceiveDeviceScopedOptions() async throws {
-        let tempRoot = FileManager.default.temporaryDirectory
-            .appendingPathComponent(UUID().uuidString, isDirectory: true)
-        let logFile = tempRoot.appendingPathComponent("sim-use-args.log")
-        let fakeSimUse = tempRoot.appendingPathComponent("sim-use")
+        let fixture = try makeFakeSimUse(versionStamp: "0.14.0")
+        defer { try? FileManager.default.removeItem(at: fixture.tempRoot) }
 
-        defer {
-            try? FileManager.default.removeItem(at: tempRoot)
-        }
-
-        try FileManager.default.createDirectory(at: tempRoot, withIntermediateDirectories: true)
-        try fakeSimUseScript(logFile: logFile.path).write(to: fakeSimUse, atomically: true, encoding: .utf8)
-        try FileManager.default.setAttributes([.posixPermissions: 0o755], ofItemAtPath: fakeSimUse.path)
-
-        let result = try await CommandRunner.run(
-            "python3 skills/sim-use/scripts/preflight.py --device target-device --sim-use-bin \(fakeSimUse.path)",
-            allowFailure: true
-        )
+        let result = try await runPreflight(fakeSimUse: fixture.executable)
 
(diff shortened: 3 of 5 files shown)


onevcat:
Thanks for this. The direction is good, and running the version gate before device discovery is the right place for it.

There is one problem: the new check rejects locally built binaries.

`sim-use --version` does not always print a plain `0.14.0`. The `VERSION` constant comes from `git describe --tags --always --dirty` (see `Plugins/VersionPlugin`), and only `scripts/local-release.sh` and `scripts/dev-install.sh` override it with `SIM_USE_VERSION`. So a binary from `make build` prints `v0.14.0` today, or `v0.14.0-3-gabc1234-dirty` on a branch.

The new regex requires no `v` prefix, and whitespace or end-of-string right after the patch number. All of those forms return `None`, so the check aborts.

I applied the patch locally and ran the script against fake binaries:

```
=== --version -> v0.14.0 ===                     <- what `make build` prints today
  FAIL  sim-use version is compatible with this skill
        hint: Upgrade sim-use to >= 0.14.0: brew update && brew upgrade ...
=== --version -> v0.14.0-3-gabc1234-dirty ===    FAIL
=== --version -> dev ===                         FAIL
=== --version -> 0.15.0 ===                      PASS
```

So anyone testing with a local build is told that their newest binary is too old. That also breaks the loop documented in CLAUDE.md (`.build/debug/sim-use describe-ui --device $UDID`).

Suggested fix: mirror what the Swift side already does. `SimUseCore/ReleaseVersion.normalize` strips the leading `v`, accepts only a clean `x.y.z`, and returns nil for dev builds. `BridgeClient` then enforces its version check only when that value is non-nil, so release builds are gated and dev builds are not. Same rule here:

```python
def parse_version(value: str) -> Optional[tuple[int, ...]]:
    m = re.fullmatch(r"v?(\d+)\.(\d+)\.(\d+)", value.strip())
    return tuple(int(p) for p in m.groups()) if m else None


def check_sim_use_compatible(ctx: Ctx) -> bool:
    result = subprocess.run(
        [ctx.sim_use_bin, "--version"],
        capture_output=True, text=True,
    )
    if result.returncode != 0:
        return False
    version = parse_version(result.stdout or result.stderr)
    # Dev and dirty stamps are unparseable by design. Only release
    # builds are gated, same rule as BridgeClient's version check.
    return version is None or version >= MINIMUM_SIM_USE_VERSION
```

## Please strengthen the tests

The current tests did not catch this, and that is the part worth fixing first. Both fake scripts print `0.14.0` / `0.13.0`, which happen to be the only two shapes the regex accepts. The tests therefore confirm the happy path rather than pin down the parsing rule, so the bug above passed CI.

Please cover the version shapes as data instead of two hand-written scripts. Concretely:

1. Make the fake `sim-use` take the version stamp as a parameter, so one helper can produce a binary for any stamp.
2. Add a parameterised test (swift-testing supports `@Test(arguments:)`) over the real shapes the CLI can emit, with the expected pass/fail for each:

   | stamp | source | expected |
   |---|---|---|
   | `0.14.0` | release build | pass |
   | `0.15.0` | newer release | pass |
   | `v0.14.0` | `make build` on a release tag | pass |
   | `v0.14.0-3-gabc1234-dirty` | `make build` on a branch | pass |
   | `dev` | no git available | pass |
   | `0.13.0` | old release | fail |
   | `v0.13.0-5-gabc1234` | old dev build | pass, if dev builds are ungated |

3. Add a case where `--version` exits non-zero, to pin the `returncode != 0` branch.
4. Keep the existing "aborts before device discovery" assertion (`log == "--version\n"`) — that part is good and should hold for every failing case.

If `parse_version` is easier to test on its own, a small Python test driven from the same Swift test would also be fine. Either way, the rule that decides pass or fail should be covered directly, not through one example per branch.

Two smaller points, not blocking:

- `MINIMUM_SIM_USE_VERSION` is a hand-written constant with nothing keeping it in sync. `scripts/local-release.sh` already gates bridge `protocol_version` and APK `versionName` parity, so a similar check or a release-checklist step would help. Otherwise the floor goes stale the next time SKILL.md documents a new command, and the gate silently stops working.
- `parse_version` is annotated `Optional[tuple[int, int, int]]`, but `tuple(generator)` is inferred as `tuple[int, ...]`.

BoWuGit:
Thanks for the detailed review — fixed in c63d1a2.

Changes:
- `parse_version` now mirrors `ReleaseVersion.normalize`: clean `x.y.z` / `vx.y.z` releases are parsed, while dev and dirty stamps remain ungated.
- a non-zero `--version` exit still fails before discovery.
- `PreflightScriptTests` now drives one shared fake binary with a parameterized 8-case matrix covering every stamp you listed, including the non-zero exit case. Every failing case retains the exact `log == "--version\n"` early-abort assertion.
- the compatibility-floor maintenance rule is now recorded beside the existing README/SKILL command-change rule in `AGENTS.md`.

Verified locally:
- `swift test --filter PreflightScriptTests` — 8 version cases plus the device-listing test pass.
- the real local binary reports `v0.14.0-1-g8a14399` and now passes the compatibility gate.
- the unrelated `ProcessControlTests.cancellableSleep` timing test passed in isolation; the full local suite intermittently exceeded its existing 5-second wall-clock threshold under load. I have not changed that unrelated test.

onevpaw (approved):
I reviewed the exact head `c63d1a2e041422b4d12b41167ab08522a0ac7d1e` and found no blocking issues. The compatibility gate runs before device discovery, correctly rejects unsupported release versions, and preserves local development and dirty builds. The implementation, tests, documentation, and changelog are consistent, and the visible CI checks pass. Thanks for addressing the initial development-build regression.

onevpaw - an assistant to @onevcat

onevcat:
@BoWuGit LGTM! Thank you for the fixes.

Issue #132, opened 2026-09-07
iOS stream-video has no native H.264 format


onevcat:
## Gap

`stream-video`'s format matrix is asymmetric:

| | iOS | Android |
|---|---|---|
| `mjpeg` / `raw` / `ffmpeg` | screenshot loop | `screencap` loop |
| `h264` | **missing** | `adb screenrecord` passthrough |
| `bgra` | raw framebuffer pixels | — |

So on iOS every practical live-stream format is a per-frame screenshot loop, which tops out around 4–5 FPS on an iPhone 17 Pro at full resolution (measured while verifying #130) and pays a codec pass per frame. Android's `h264` leg streams at the device's native variable frame rate with no per-frame encode on our side.

## The plumbing already exists

`record-video` has driven native H.264 on iOS since #52/#70:

```swift
let config = FBVideoStreamConfiguration(
    format: .compressedVideo(withCodec: .h264, transport: .annexB),
    framesPerSecond: fps,
    rateControl: .quality(Double(quality) / 100.0),
    scaleFactor: scale,
    keyFrameRate: 2.0
)
```

`record-video` points that at a file. `stream-video --format h264` would point the same configuration at stdout, which is what the Android leg already does with `screenrecord`'s Annex B output — and `SimUseVideo` already has the host-side Annex B parsing from the recording work.

This would also close the "h264 is Android-only for now" caveat in the 0.12.0 changelog and make the `stream-video` format set symmetric across platforms.

## Scope notes

- `--fps` semantics: iOS `record-video` honours a constant `--fps`, so the stream can too (unlike Android's variable rate).
- `streamBGRA` already streams from `FBVideoStream` and handles startup/mid-stream failure detection (#21); that error taxonomy is the model to follow.
- Worth checking whether `ffplay -f h264 -probesize 32 -fflags nobuffer -` (the pipeline the README documents for Android) works unchanged against the iOS output.

onevcat:
## Spike results: `h264` is the viable native format; `.mjpeg` is not

Ran the experiment on a booted iPhone 17 Pro / iOS 27.0. idb turns out to expose more than I wrote above — the full format enum is:

```swift
// FBControlCore/Configuration/FBVideoStreamConfiguration.swift
public enum FBVideoStreamFormat: Hashable {
  case compressedVideo(withCodec: FBVideoStreamCodec, transport: FBVideoStreamTransport)
  case mjpeg      // raw JPEG bytes, unframed
[...]

PR #133, opened 2026-09-07
feat: unify iOS record and stream onto one H.264 capture path

onevcat:
Closes the iOS half of #132.

## Summary

`stream-video` had no native format on iOS — `mjpeg`/`raw`/`ffmpeg` were a screenshot-per-frame loop, `bgra` was raw pixels — so watching a simulator live meant ~4 fps at ~1.9 MB/s. `record-video` has driven native H.264 since #52/#70, so the capability was already there; it just wasn't wired to stdout.

Both verbs now build one `FBVideoStreamConfiguration.h264Capture(fps:quality:scale:)` and differ only in their sink:

```
record-video  ->  H264MuxingPipeline  ->  MP4
stream-video  ->  stdout
```

That muxer's own doc comment already described it as *"shared by the iOS (`FBSimulatorVideoStream`) and Android (`adb screenrecord`) capture paths"* — iOS simply never connected to it, using idb's separate in-process file writer instead. `AnnexBPipelineConsumer` is the ~25-line adapter from idb's byte-stream consumer protocol to that pipeline.

## Numbers

Booted iPhone 17 Pro / iOS 27.0, 6-second captures:

| path | frames | effective fps | bytes |
|---|---|---|---|
| screenshot loop (`mjpeg`) | 24 | 4.1 | 11.2 MB |
| screenshot loop (PNG) | 26 | 4.5 | 92.8 MB |
| **native h264 Annex B** | **144** | **24.0** | **1.33 MB** |

~5.9x the frame rate for an eighth of the bytes.

## The one real trade-off

Annex B carries no container timing, so frame timestamps now come from host arrival time rather than the encoder's sample clock. Three runs of each, under real touch input:

| path | inter-frame jitter (stddev) | measured fps |
|---|---|---|
| new (`annexB` → shared muxer) | 6.0 / 6.4 / 6.8 ms | 30.3 |
| old (idb in-process writer) | 3.1 / 2.7 ms | 29.8 |

Roughly 2x the jitter, but both are far inside a 33 ms frame at 30 fps — no dropped or reordered frames, and playback is indistinguishable. The `--fps` constant-rate assertion and the 100 ms-grace SIGTERM test (5 iterations, each requiring a valid MP4) both still pass.

## Why not fmp4

Tried first, rejected on evidence. `FBFMP4FrameWriter` emits live fragments with `mvex` but no `sidx`/`mfra`, and AVFoundation needs an index to treat fragments as a sample table:

| file | AVFoundation (QuickTime / Finder / `GIFTranscoder`) | ffmpeg |
|---|---|---|
| idb live fMP4 | **duration 0.000s, 0 frames** (unchanged by `PreferPreciseDurationAndTiming`) | 133 frames |
| same data, ffmpeg-remuxed | 133 frames | 133 frames |

A recording would have been unplayable on double-click and untranscodable to GIF. Annex B keeps the MP4 a regular non-fragmented file.

## Verification

- `make build`, `make test` — 1385 pass
- Live on iPhone 17 Pro / iOS 27.0: `ios stream-video --format h264`, the top-level `stream-video --format h264` on an iOS UDID (previously rejected as Android-only), `ffmpeg -f h264` remux to 1206x2622, `bgra` non-regression (byte-identical size to before), `record --format gif` (29 frames)
- iOS E2E: **`StreamVideoTests` 9/9**, **`RecordVideoTests` 7/7** — including the constant-`--fps` assertion and the repeated short-grace SIGTERM finalize test
- Android untouched, and no emulator was reachable here, so no Android E2E was run

The `Stream video rejects invalid formats` E2E asserted on `--format h264` being invalid; it now uses `webm`, a value no backend defines.

## Follow-up

Deprecating `mjpeg`/`raw`/`ffmpeg` is deliberately **not** in this PR. It's an API change, and the Android side needs `stream-video` to gain the automatic `screencap` fallback that `record-video` already has before those formats can stop being user-selectable — which needs an emulator to verify. Filing separately.


onevcat:
Pushed a follow-up fix after live testing exposed a defect the automated checks could not see.

**Bare Annex B accumulates unbounded latency.** The stream carries no presentation timestamps, so a player guesses the frame rate — ffprobe reads ours as 25 fps regardless of `--fps` — and a 30 fps capture consumed at 25 falls ~5 frames further behind every second. Watching the window while driving the simulator, the picture ended up minutes behind the device. Short automated captures never surface this: the drift needs sustained viewing to become visible, and my earlier ~6 s measurements looked fine.

`--format h264` now emits H.264 in MPEG-TS, which carries PTS/DTS on a 90 kHz clock:

| | annexB | mpegts |
|---|---|---|
| `avg_frame_rate` | `25/1` (guessed) | `30/1` |
| `start_time` | N/A | 0.0028 |
| `duration` | N/A | 4.30 s |
| ffplay `vq` | 293 KB and rising | **0 KB across a 60 s run** |

`transport` became a parameter of the shared capture factory instead of being baked in, since the two sinks genuinely need different framing: the file sink keeps `.annexB` (`H264MuxingPipeline` parses start codes), the stdout sink takes `.mpegts`. Still one configuration and one code path — the sink picks the transport.

The h264 E2E now pins 188-byte MPEG-TS packet alignment rather than an Annex B start code, and the preview command in both the README and the command's own stderr hint is `ffplay -f mpegts`.

**Android has the identical flaw**, verified on a fresh emulator (`avg_frame_rate=25/1`, `start_time=N/A`). It is not fixed here because its stream is a `screenrecord` passthrough we never encode — `--output-format` has no TS option, so fixing it means host-side re-containering. Filed as #135; this PR documents the per-platform container and the re-container one-liner as an interim measure.

## Verification since the last push

- unit 1385 pass; CI green
- iOS E2E: `StreamVideoTests` **10/10**, `RecordVideoTests` **7/7**
- **Android E2E: 10/10 suites** on a freshly created API 36 emulator (`AndroidStreamVideoTests` and `AndroidRecordVideoTests` included) — confirms this PR leaves the Android paths alone
- 60 s live viewing with sustained interaction: queue stays empty, picture tracks the device

PR #136, opened 2026-09-07
fix: carry the Android h264 stream in MPEG-TS so players stay in sync

### LINKED ISSUE #135: Android stream-video --format h264 accumulates unbounded latency (bare Annex B has no PTS)
Split out of #132 / #133, which fixed the same flaw on iOS.

## Problem

`sim-use android stream-video --format h264` passes `adb screenrecord --output-format=h264` through byte-for-byte. That output is bare Annex B, which carries no presentation timestamps:

```
$ sim-use android stream-video --format h264 --device emulator-5554 > out.h264
$ ffprobe out.h264
codec_name=h264
r_frame_rate=1200000/1
avg_frame_rate=25/1      <- guessed, not real
format_name=h264
start_time=N/A
duration=N/A
```

A player has to pick a frame rate, and ffmpeg's default guess is 25 fps. When the real rate is higher, the consumer drains slower than the producer fills and the lag grows without bound — the picture ends up minutes behind the device. This was observed directly on the iOS leg before it moved to MPEG-TS; Android's stream has identical metadata, so it has the same failure mode.

Archiving is unaffected — `ffmpeg -f h264 -i - -c copy out.mp4` re-stamps the timeline on the way into the container. It's live viewing that degrades.

## Why the iOS fix doesn't port

iOS could simply ask idb for a different transport (`FBVideoStreamTransport.mpegts`), because sim-use drives that encode. Android's frames come out of `screenrecord` already encoded; `--output-format` offers `h264`, `webm` and `mp4`, and only `h264` is streamable (`mp4` writes a trailing `moov` and so must be a complete file). There is no TS option to ask for.

## Options

1. **Host-side re-containering.** Wrap the Annex B in MPEG-TS on the way to stdout, stamping PTS from host arrival time — exactly what `H264MuxingPipeline` already does for the file sink, minus the MP4. Keeps the CLI contract, makes live preview work, costs a TS muxer (or shelling out to ffmpeg, which we currently don't depend on at runtime). Note the device's rate is genuinely variable, so host arrival time is the honest timeline here, not a constant `--fps`.
2. **Document only.** State that Android's `h264` is for archiving and give the re-container one-liner (`… | ffmpeg -f h264 -i - -c copy -f mpegts - | ffplay -f mpegts -`). Zero code, but the obvious ffplay invocation stays a trap.
3. **Reuse the muxing pipeline for a `--format mp4` file sink** and leave `h264` as the archival raw stream.

#133 took option 2 as an interim measure — the README now names the container per platform and points at the re-container command.

## Repro

```bash
sim-use android stream-video --format h264 --device <serial> | \
  ffplay -f h264 -probesize 32 -fflags nobuffer -
# interact with the device; the window falls progressively further behind
```

onevcat:
`screenrecord --output-format=h264` emits a bare elementary stream, which
carries no presentation timestamps. A player has to guess a frame rate —
ffprobe reads ours as 25 fps whatever the device really does — and since
the device commonly encodes faster, the consumer drains slower than
sim-use fills. The lag then grows without bound rather than settling.
Measured with ffplay while driving the device:

  elapsed   queued video
      5 s   0 KB
     10 s   0 KB
     20 s   578 KB
     30 s   1376 KB

Now re-containered into MPEG-TS on the way to stdout, matching the iOS
surface: one `ffplay -f mpegts` invocation works on both platforms. The
same run measures 0 KB queued throughout. Frames are never re-encoded;
only the wrapper is added.

`MPEGTSMuxer` is a minimal one-program muxer — PAT, PMT, and PES with a
90 kHz PTS — reached through the existing `H264MuxingPipeline`, whose sink
is now the `H264AccessUnitSink` protocol so a file recorder and a stream
writer can share the Annex B parsing and timing work.

Two behaviours fell out of the change and are handled deliberately:

- Program tables are written before the first picture and re-announced
  every 0.5 s. A consumer attaching mid-stream cannot decode until it has
  seen them, and on a variable-frame-rate source the first frame may never
  come (a still screen produces none at all).
- That re-announcement doubles as the keep-alive that detects a hung-up
  consumer. The old byte-passthrough noticed a closed pipe because it was
  always writing; a re-containering path on a still screen writes nothing,
  and without the tables the process ran on indefinitely after the reader
  went away. It now exits in 0.5 s.

Fragmented MP4 was tried first and abandoned. AVAssetWriter's HLS
segmentation wants to cut fragments on keyframes, and `screenrecord` emits
them too rarely to deliver anything timely — an 8 s capture came out as a
single 15 MB fragment handed over only at stop, which is a file, not a
stream. Driving `flushSegment()` manually deadlocked against the writer
lock. A transport stream has no keyframe constraint, every PES packet
standing on its own, which is why broadcast and HLS use it.

Verified: unit 1393 pass (8 new muxer conformance tests covering packet
framing, continuity counters, PTS marker-bit layout, CRC-32/MPEG-2, and
adaptation fields); Android E2E 10/10 suites; live output demuxes to 314
frames with `duration=7.36 s` and delivers continuously (2.9 MB by 2 s,
16.7 MB by 8 s) where the fMP4 attempt delivered nothing until stop.

Refs #135

Signed-off-by: onevcat <onevcat@gmail.com>

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

---

## Verification

- unit **1393 pass** — 8 new muxer conformance tests (packet framing, continuity counters, PTS marker-bit layout, CRC-32/MPEG-2, adaptation fields, parameter-set repetition)
- **Android E2E 10/10 suites** on a headless API 36 emulator
- Live output: `format_name=mpegts`, `duration=7.36 s`, 314 frames, decodes without warnings; 89,799 packets all correctly sync-byte aligned
- Continuous delivery (the property fMP4 lacked): 2.9 MB by 2 s, 7.0 MB by 4 s, 11.5 MB by 6 s, 16.7 MB by 8 s
- Consumer hangup now exits in 0.5 s (it hung indefinitely at one point during development, on a still screen — see the keep-alive note above)

Stacked on #133.

Closes #135.


onevcat:
**Note on CI:** this repo's workflows trigger on `pull_request: branches: [main]`, so a stacked PR whose base is another branch gets only the DCO check. Ran the full set locally instead:

- `make test` — **1393 pass**
- `make e2e-android` — **10/10 suites** (headless API 36 emulator)
- iOS E2E `StreamVideoTests` + `RecordVideoTests` — **both suites pass**, which is the one that matters here: this PR turns `H264MuxingPipeline`'s sink into a protocol and widens `makeSampleBuffer`, both of which iOS `record-video` depends on.

CI will run in full once #133 merges and this rebases onto `main`.

onevcat:
## Round 1 adversarial review — all findings accepted and fixed

An independent reviewer went at the muxer against ISO/IEC 13818-1. Every finding was real; a passing test suite and a working ffplay preview were hiding all of them. Each fix landed with a test that fails first.

### [P1] PSI sections were malformed — accepted, fixed

The byte after `table_id` carries `section_syntax_indicator` plus two reserved bits; both were zero. The PMT also declared length 17 where its fields sum to 18. A receiver that reads PSI properly could not associate the video with the program.

Confirmed with the reviewer's own probe, before → after:

| | before | after |
|---|---|---|
| `nb_streams` | **0** | 1 |
| `pcr_pid` | **0** | 256 (0x100) |
| `codec_name` | (heuristic scan only) | h264 |

Now `0xB0, 0x0D` / `0xB0, 0x12`. New tests parse the emitted sections — indicator, reserved bits, exact length, CRC residue over the whole section, PCR PID, `stream_type`, ES mapping — rather than only asserting which PID they arrived on.

### [P1] PCR-only packet shape — accepted, fixed

It signalled `adaptation_field_control = 0b11` while stuffing away all 184 bytes, and advanced the PID's continuity counter. 13818-1 §2.4.3.3 excludes payload-less packets from the counter, so that made receivers report loss that never happened. Now `0b10`, counter untouched.

One correction to the finding: "do not advance the counter" means the packet *repeats the current value* and the next payload packet consumes it — not that it repeats the previous packet's value. My first test asserted the latter and was wrong; the test now encodes the standard's rule.

### [P1] Concurrent producers could interleave the stream — accepted, fixed

Correct, and the consequence was worse than described. Generation and delivery are now one critical section.

The reviewer suggested a dedicated serial writer. I used the existing lock instead: the ordering that matters is exactly the ordering of muxer state mutation, so widening that critical section to cover the write gives the guarantee without a second synchronisation primitive to reason about. A blocking write now holds the lock and the capture thread waits — that is the backpressure working, and it is the same shape `record-video` already has.

The new test measures the property instead of racing for it — a sink counting concurrent `consume` calls. It saw **2 before, 1 after**. (My first attempt used a task group; two synchronous child tasks with no suspension point ran sequentially and proved nothing.)

### Two further defects found while verifying — not in the review, fixed

Chasing a `Packet corrupt` warning that appeared **once per picture** on a live stream. I built an offline harness that muxes a captured elementary stream, which isolated both causes from the device:

1. **`PTS == PCR`.** A receiver presents a picture when its clock reaches the PTS; equal values leave no time to decode. PTS now leads by a 100 ms decode allowance — a constant timeline shift, not a per-picture cost.
2. **Keep-alives restated the previous PCR.** A PCR samples the transmission clock and must advance, but this timeline only advances when a picture arrives, so during an idle stretch there is no honest value to send. Interleaving keep-alives into the offline mux produced one warning per picture; removing the clock from them produced none. Keep-alives now announce structure only; each picture carries its own PCR.

Also widened the minimum step between pictures arriving in one chunk from a single 90 kHz tick to 1 ms — a tick collapses to the same value in a coarser downstream timebase, which ffmpeg rejected as non-monotonic when muxing to a file.

**A live Android stream now demuxes with no ffmpeg warnings of any kind.** Previously: one corrupt-packet warning per picture.

### Still open from round 1

- **[P2] `stream-video | ffmpeg -c copy` yields a compacted timeline.** Valid — the 0.2 s cap turns a 60 s idle into 0.2 s of output, and the README presents that command without qualification. Addressing next.
- **[P2] "one capture path" overstates it.** Also valid: `record-video` takes Annex B through the host muxer, `stream-video` takes idb's MPEG-TS to stdout. They share encoder settings, not a transport. Wording to be corrected.

Verification: unit **1402 pass**, Android E2E **10/10**, live stream clean under `-loglevel warning`.

onevcat:
## Round 2 — both P1s accepted and fixed, one of them mine

### [P1] A stalled consumer made the command unkillable — accepted, fixed

**This was introduced by round 1's own fix.** Moving the write inside the muxer's critical section guaranteed byte order and left a blocking write holding the lock. The reviewer identified the case that a closed pipe does not cover: a consumer that keeps the pipe open but stops draining never produces EPIPE — the pipe just fills. `Darwin.write` then blocked with the lock held, no blocked write ever consulted the cancellation flag, and the command loop could not reach `process.interrupt()`. Ctrl-C needed SIGKILL, with adb left running.

stdout is now non-blocking; a full pipe surfaces as EAGAIN and the sink waits in 100 ms slices, checking cancellation between them. Cancellation and a broken pipe became distinct outcomes — only the latter latches the sink closed, so an interrupted write is no longer misreported as consumer hangup.

The sink now takes its file descriptor, which is what made this testable: a pipe nobody drains, a write on another thread, then cancellation. **Mutation-checked** — deleting the cancellation check makes the test fail with `write never returned after cancellation`, so it holds the property rather than describing it.

I did not build the dedicated serial writer with bounded queueing that was suggested. Backpressure that blocks the capture thread is the behaviour I want — it is what `record-video` already does, and buffering instead would mean either unbounded memory or dropping pictures. What was actually broken was that the block could not be interrupted, and that is now fixed directly.

### [P1] The compacted timeline was undeclared — accepted, fixed

The reviewer was right on both counts: an advancing PCR cannot coexist with a compacted PTS timeline, and freezing the clock instead is not a coherent model. What we both missed is that the format has a bit for exactly this — **ISO/IEC 13818-1 §2.4.3.4 `discontinuity_indicator`**.

A picture resuming after an idle stretch now declares it. That turns a clock that jumps from a defect into a documented transport discontinuity: a receiver disciplining its own clock from the PCR resynchronises rather than treating the picture as wildly late. Set only when the real gap exceeded the cap, so a normal inter-frame gap does not make a receiver resynchronise on every picture.

This leaves the remaining objection — PCR repetition exceeding 100 ms while idle — as a deliberate and now-declared property rather than an accident. Sending a clock during an idle stretch is the one option that is closed off: this timeline only advances when a picture arrives, and repeating the previous value is precisely what produced a corrupt-packet warning per picture in round 1.

On the correction I asked about: confirmed against §2.4.3.3 — a payload-less packet repeats the *current* counter value and the next payload packet consumes it. My original test asserted it repeats the previous packet's value, which was wrong.

### [P2] Docs promised idle clock references — accepted, fixed

Correct, and it was concealing the P1. The README and changelog said idle output carries "program tables and clock references"; it carries tables only. Both corrected, along with a third stale mention in the preview-command entry.

## Verification

Live on a freshly restarted emulator, driving it with an idle stretch in the middle:

| | |
|---|---|
| pictures | 167 |
| PCR samples | **167** (one per picture) |
| discontinuities declared | **2** (matching the idle stretch) |
| program association | `nb_streams=1  pcr_pid=256  codec_name=h264` |
| ffmpeg `-loglevel warning` | **no warnings of any kind** |

Unit **1406 pass**; Android E2E **10/10**.

One process note: an earlier push of this commit briefly contained CHANGELOG conflict markers — I ran `git add -A` while a stash pop was still unresolved. Caught, amended, and all three layers verified marker-free before this push.

onevcat:
## Round 3 — one finding accepted, one rejected on measurement

### [P2] The recorder described an architecture this stack replaced — accepted, fixed

`H264PassthroughRecorder`'s comment said `adb screenrecord` was the "only remaining caller" and that iOS recorded through idb's own file writer. iOS now constructs this same recorder, so the comment was actively misleading for anyone touching timestamping.

Restated as the invariant that actually holds, per the reviewer's suggestion: an Annex B elementary stream carries no per-access-unit timing, so arrival time is the only clock available — whichever capture produced the bytes. That does not go stale again.

### [P1] Replace the capped timeline with a wall-clock one — rejected, measured

The argument was that an advancing PCR lets a receiver hold the last picture while idle and present the next immediately, so both the cap and the discontinuity flag could be deleted. That would have been a genuine simplification, so I built it: one wall-clock timeline, PCR sampled every 40 ms through the idle stretch, compaction and discontinuity removed. Unit tests rewritten for the new contract and passing — including one asserting no PCR gap exceeds 100 ms.

Then I measured it on the emulator:

| idle | wall-clock timeline (proposed) | capped timeline (current) |
|---|---|---|
| 10 s | **had not caught up 30 s later** | 0.45 s |
| 30 s | **had not caught up 30 s later** | 0.51 s |

**The premise does not hold for the players this stream feeds.** ffplay schedules presentation from the pictures' own timestamps — with no audio its master clock is driven by the last picture it showed — so an advancing PCR does not move it. The next picture, stamped with the time it really arrived, is due far in that clock's future. ffmpeg and browser MSE schedule from PTS the same way. An advancing transport clock only helps a receiver that disciplines its STC from the PCR and presents against that, which is not what is consuming this.

So the deviation stands, and the code now carries the measurement instead of an assumption (`MPEGTSStreamWriter.swift`, on `maxFrameGap`). The comment also states plainly the part the reviewer is right about: **PCR repetition exceeds the 100 ms limit while the screen is still**, because a timeline that only advances on a picture has no honest clock value to send in between — and restating the previous one is exactly what produced a corrupt-packet warning per picture in round 1. Acceptable for a preview consumed by PTS-scheduling players; it would not be acceptable in a broadcast mux. The resuming picture declares a transport discontinuity so a clock-disciplining receiver at least knows to resynchronise rather than treating it as late.

I also did not adopt the suggestion that `keepAliveCarriesNoClock` "codifies the invalid case". It pins the behaviour that a measured alternative is worse than — which is what a test is for. Its comment explains why rather than asserting the standard's preference.

### Simplifications applied

- `MPEGTSMuxer.headerSize` — declared, never read. Removed.
- Two identical latches in `StdoutStreamSinkTests` collapsed into one.
- The capture factory's comment cut to its local contract (why the two sinks need different `transport` values), dropping rationale already documented in the muxer and README.

Left in place as advised: the write-ordering rationale and the adaptation-only continuity note. Both record constraints invisible from the code, and bugs this stack actually hit.

## Verification

- unit **1406 pass**
- **Android E2E 10/10**; iOS E2E `StreamVideoTests` + `RecordVideoTests` both pass (this stack changes shared `SimUseVideo` code that iOS recording depends on)
- live stream: 142 pictures, `nb_streams=1  pcr_pid=256  codec_name=h264`, ffmpeg demuxing with **no warnings of any kind**

## Where this leaves the review

Three rounds, nine findings, all real. Fixed: PSI section syntax, PCR-only packet shape, concurrent write ordering, uninterruptible blocked write, PTS/PCR equality, keep-alive clock restatement, two documentation claims the code did not support, and a stale architecture comment. One rejected with measurement. Two of the defects were introduced by earlier rounds' own fixes, which is the argument for having done more than one round.

PR #141, opened 2026-09-11
Build the Android backend on Linux

iopass4:
## Summary

Builds the Android half of sim-use on Linux. The Android backend drives devices through `adb` and the bridge APK's HTTP server and needs
 none of the Apple frameworks, so this adds an `#if os(Linux)` target graph — `SimUseCore`, `AndroidBackend`, and a new `SimUseLinux` ex
ecutable — next to the unchanged macOS manifest. The main use case is agents running in Linux containers, CI, and WSL.

**The macOS build is unchanged:** the macOS manifest sits verbatim in the `#else` branch, and every source change is either compiled out
 on Apple platforms or behaviour-preserving (details below, verified with `make test` and on a simulator).

## Commits (review one at a time)

1. **fix: write the outline cache with rename(2) on Linux** — swift-corelibs-foundation's `replaceItemAt` fails whether or not the targe
t exists, so `ui` → `tap @N` broke on Linux. Apple platforms keep `replaceItemAt`. Adds an overwrite test.
2. **fix: report a missing adb binary as `adbMissing` on Linux** — corelibs reports it as Cocoa 260, not 4 / ENOENT. Already covered by
`testRunMapsMissingBinaryToAdbMissing`.
3. **build: make SimUseCore and AndroidBackend imports portable** — conditional `Darwin`/`Glibc`/`os`, `FoundationNetworking`; an `inter
nal` `LinuxCompat.swift` (compiled out on Apple platforms) provides a `Darwin` namespace forwarding to Glibc, so the daemon's qualified
`Darwin.close(fd)` calls compile unchanged, plus an `NSLock`-backed `OSAllocatedUnfairLock`.
4. **refactor: move the daemon command into SimUseCore** — the command only needed the executable for three hooks (root parser, iOS stal
e-HID cleanup, liveness probe); they are now injected via `Daemon.installPlatformHooks`, called at the same point in `daemon start`. Avo
ids a second copy for the Linux executable.
5. **feat: accept `ui` on android describe-ui** — matches the top-level / iOS verb; the Linux root depends on it for `sim-use ui`.
6. **build(linux): add an Android-only Linux target graph** — the manifest branch, `SimUseLinux`, and `AndroidCommand` registering recor
d/stream only when `SimUseVideo` is importable (macOS subcommand order unchanged).
7. **feat: reach the bridge through a remote adb server** — `adb forward` listens on the adb server's host; when `ADB_SERVER_SOCKET=tcp:
<host>:<port>` is remote (e.g. WSL → Windows adb), the bridge is reached there. `SIM_USE_BRIDGE_HOST` overrides. Pure function + tests (
IPv6/zone ids). Also applies on macOS; the default (no env) is unchanged.
8. **docs** — `docs/linux.md` (scope, install, WSL setup incl. the security implications of `adb -a`), README, AGENTS.md, skill, CHANGEL
OG.
9. **ci** — a `swift:6.3` ubuntu job running `swift build` / `swift test` for the Linux graph.

## Scope on Linux

Available: the Android verbs at the top level (`ui`, `tap`, `type`, `paste`, `swipe`, `button`, `touch`, `gesture`, `multi-touch`, `keyb
oard-state`, `screenshot`, `devices`), the full `android` namespace, and the per-device daemon.

Not available: iOS verbs, `record-video` / `stream-video` (AVFoundation), `long-press` / `app-state` (their Android paths live in the ma
cOS forwarders, which import iOS modules), the Viewer, and `init`. `sim-use devices` is the Android listing rather than the unified sche
ma. Bringing `long-press` / `app-state` / unified `devices` over would mean moving their Android halves into `AndroidBackend` — happy to
 do that as a follow-up if you want it.

## Testing

- **Linux** (Swift 6.3.3 on WSL2 Ubuntu, and the `swift:6.3` image from a clean `git archive`, i.e. without the bridge APK — the same th
ing the new CI job runs): `swift build`, `swift test` — 262 XCTest + 133 swift-testing, all green. Commits 6 and 7 also build and pass o
n their own.
- **macOS** (macOS 26.3, Xcode 26.6): `scripts/build.sh dev`, `make build`, `make test` — 311 XCTest and 1106 swift-testing tests pass.
One test, `ProcessControlTests` "cancellableSleep wakes early when the flag is cancelled", failed in 2 of 3 full runs (5.6 s / 9.6 s aga
inst its 5 s bound). It fails the same way on `main` on the same machine (5.6 s) and passes 5/5 in isolation, so it looks like a pre-exi
sting timing flake under full-suite load rather than something this PR introduces.
- **macOS, live**: on an iPhone 17 simulator (iOS 26.5), `ui` spawns the daemon through the new `installPlatformHooks` path, and its out
put is identical to the `main` binary's; `daemon status` / `daemon stop --all` work as before.
- **Linux, live**: against a real Android device over network adb — `android init`, `android ping`, `ui` twice (cache created, then over
written), and `tap @N` resolving from the cache; `daemon start` / `status` / `stop` through the shared command; `scripts/install-linux.s
h` into a scratch prefix.


onevcat:
Thank you for your contribution! 

That is actually what on our plan list and the implementation looks good for the first glance. I will try to take a deeper look soon and see if we need more modification before we can determine the next action.

I will reach you back soon.

onevtail (changes requested):
The Linux cache fix is supported by the reported reproduction and concurrent read/write checks. Review of head `27b558c09c785b6940c592d3469ad1f23d2ddc76` identified two blocking issues in remote adb connection scoping:

1. **[P1] A warm daemon keeps targeting the previous adb server.** [BridgeClient.swift](https://github.com/lycorp-jp/sim-use/blob/27b558c09c785b6940c592d3469ad1f23d2ddc76/Sources/AndroidBackend/Bridge/BridgeClient.swift#L358) derives its host from the daemon's environment, while daemon lookup uses only the device serial and forwarded requests carry no connection configuration. Starting with server A and then switching the client to B with the same serial returned `ok:true` for both `button home` commands, but both button requests reached A. Please include the connection configuration in daemon identity or validate it before reuse; recalculating the host inside the existing daemon would still read its unchanged environment.

2. **[P1] Restarting the daemon can send the previous session's credentials to the new host.** [Session restoration](https://github.com/lycorp-jp/sim-use/blob/27b558c09c785b6940c592d3469ad1f23d2ddc76/Sources/AndroidBackend/Bridge/BridgeClient.swift#L93) restores the port and token by serial, allowing the [URL construction](https://github.com/lycorp-jp/sim-use/blob/27b558c09c785b6940c592d3469ad1f23d2ddc76/Sources/AndroidBackend/Bridge/BridgeClient.swift#L398) to combine B's host with A's cached port and token. With A using port 18080 and B expected to use 18081, the cold-start probe reached an unrelated mock listener at B:18080 carrying A's synthetic token. Reconnecting only after a connection failure does not cover an old port that still responds. Please persist and validate the session's connection identity, and recreate forwarding and credentials when it changes or is absent from a legacy cache.

Please add regression coverage for switching servers with a warm daemon, switching with a persisted session, and an unrelated listener occupying the cached port. These reproductions used simulated adb/HTTP services and synthetic credentials; real Android/WSL end-to-end behavior remains unverified.

onevtail - an assistant to @onevcat

iopass4:
Thanks for the careful repro. I could reproduce both issues. The fixes are in three commits on top of 27b558c:

**e3c7365, scope cached connections to the adb server**
- Warm daemon: the daemon reports the connection it was started with (adb server + bridge host) in `_ping`, and the client restarts the daemon when its own connection is different.
- Persisted session: `bridge.json` stores the connection too. A cached session is reused only if the connection matches and `adb forward --list` still shows its port for that serial. Otherwise the forward and token are recreated. Old cache files without the field are ignored.

**8175e51, two gaps in the first fix**
- The connection check now applies to every adb serial, including long mDNS serials that `looksLikeAndroid` rejects but that still get a daemon.
- If `adb forward --list` fails, the error is returned and the cached session is kept, instead of opening another forward.

**4583430, found while testing on a real device**
- Removing a stale forward now passes `-s <serial>`. Without it, adb rejects the command when the server has more than one device, so old forwards piled up.
- `android init` prints the actual bridge host instead of always `localhost`, and `--json` includes `bridgeHost`.

**Tests**
The three cases you asked for are covered: switching servers with a warm daemon (`DaemonConnectionGateTests`), and switching with a persisted session plus an unrelated listener on the cached port (`BridgeConnectionScopingTests`).

**Verification**
- Your scenarios against the real binary, with fake adb servers and mock listeners: before the fix, B's requests went to A, or to the unrelated listener with A's token. After it, they reach B with B's token.
- A real phone from WSL, with the same serial connected to both the Windows adb server and WSL's own adb server: with 27b558c, a warm daemon kept using the first server after the client switched. With the fix, the daemon restarts and the forward is created on the new server, both with a warm daemon and from a persisted session. A live cached forward on the same server is still reused.
- `swift test` passes on Linux. On macOS (Xcode 26.6), `make build` passes, and `make test` passes except for the `ProcessControlTests` "cancellableSleep wakes early" timing check, which also fails intermittently on `main`.

iopass4:
hi,
   The macOS failure is `ProcessControlTests.swift:42` (cancelled sleep took 6.2s, limit 5s). This test is already flaky upstream: it failed the same way on main at 397dcb0 ([run 35058898057](https://github.com/lycorp-jp/sim-use/actions/runs/35058898057)). This PR doesn't change that code on macOS. It only wraps `import os` in `#if canImport(os)`.

onevcat:
@iopass4 

Yes! The failing test is unrelated to this PR.

I’ll try to find some time to go through this in detail and see how it all works. Thank you for the great work—I really appreciate it!

Issue #142, opened 2026-09-15
Coordinate-space contract differs across tap/touch/swipe/gesture, and is only partly documented


iXerol:
tap: no `--coordinate-space` opt-in for raw coordinates, and the per-verb contract is undocumented

## Context

Raw coordinates mean different things to different verbs, and the difference only shows up on a rotated device. Measured on sim-use 0.14.0, iPhone 17 / iOS 27 simulator, Xcode 27.0 (27A266a):

| Verb / form | Raw-coordinate space | Opt into visual space | In the skill |
| :-- | :-- | :-- | :-- |
| `swipe` | device-native portrait | `--coordinate-space ui` | pitfalls row |
| `touch` | device-native portrait | `--coordinate-space ui` | same row ("on `swipe`/`touch`") |
| `tap -x/-y`, `tap --point` | device-native portrait | **no such option** | — |
| `tap @N` / `#id` / `--label` | n/a — orientation-calibrated | n/a | Act table |
| `gesture` single-finger presets | **visual space**, orientation-aware | n/a | three examples only |
| `gesture` pinch / rotate presets | device-native portrait | no | three examples only |

`sim-use ios tap --help` lists no `--coordinate-space`; `swipe --help` and `touch --help` both do. In `IOSSimTapCommand.execute()` the explicit-coordinate branch sets `calibration = nil` unconditionally, while the selector branches take `hidTarget.calibration` — so a selector tap follows a rotation and a coordinate tap does not, with no flag to change that.

The two `gesture` families disagree with each other, which `gesture --help` states plainly ("Single-finger presets: visual space, auto-detected from the current orientation… Pinch/rotate presets: device-native portrait space") but nothing else does.

This is the shape that bites an agent: `ui` prints visual-space frames, the natural next step is `tap -x/-y` with a number read off that outline, and on a rotated device the tap lands somewhere else while the command reports success.

## Proposal

Give `tap` the same `--coordinate-space` option `touch` already has, so the raw-coordinate path behaves uniformly across the three HID verbs.

- **Where**: `IOSSimTapCommand` — declare `@Option(name: .customLong("coordinate-space")) var coordinateSpace: CoordinateSpace = .native`, and in the `TapCoordinateResolver.resolve` branch replace the hardcoded `calibration = nil` with a `UISpaceCalibrationLoader.load` call when the value is `.ui`, mirroring `IOSSimTouchCommand`.
- **Default stays `native`**, so no existing caller changes behaviour.
[...]

PR #144, opened 2026-09-16
fix: give tap the --coordinate-space opt-in swipe and touch already have

iXerol:
## Summary

#74 added `--coordinate-space` to `swipe` and `touch`, but not to `tap`. So
the one verb an agent reaches for after reading `describe-ui` was also the one
that could not consume the outline's own coordinates: `tap -x/-y` always
dispatched in device-native portrait space, and on a rotated device it landed
somewhere else, with no flag to correct it.

```bash
# landscape-right device; the outline prints the element at (121,28 71x36)
sim-use tap -x 156.5 -y 46                          # dispatches (156.5, 46.0) — misses
sim-use tap -x 156.5 -y 46 --coordinate-space ui    # dispatches (356.0, 156.5) — hits
```

`ui` runs the point through the same `OrientationCalibration` that selector
taps have ridden since #34. `native` stays the default and stays zero-cost —
no AX round-trip, no calibration — so no existing caller changes behaviour.
Batch `tap` steps take the batch-wide calibration, the way `swipe` / `touch`
steps already do.

## The one decision that wasn't already settled

`tap` is the only verb that takes **both** coordinates and selectors, so it is
the first place `--coordinate-space` can be combined with a target that does
not have user-supplied coordinates. On `swipe` / `touch` the combination is
structurally impossible, so #74 never had to answer it.

Aliases and selectors read their coordinates from the AX tree — already ui
space, already calibrated — so the flag has nothing to act on there. It
rejects rather than silently accepting a no-op:

```
$ sim-use tap --label 'OK' --coordinate-space ui
Error: --coordinate-space ui applies to explicit -x/-y/--point coordinates only;
aliases and selectors are already resolved in ui space.
```

That follows the precedent `--frame` set on this same command, which already
rejects the targeting forms it cannot apply to (`--frame` + explicit
coordinates, `--frame` + `@N`). Silently ignoring it was the alternative, and
is what Android does — but there the whole flag is a no-op, so there is
nothing to mislead anyone about.

`android tap` gains the flag as the same documented no-op `android swipe` /
`android touch` already carry, closing the last parity gap in that family.

## Verification

`make test` — **1418 passed, 0 failed**, no new warnings. 22 of those are new
or extended, across five suites: the tap opt-in validation table (explicit
coordinates accept, selector and alias reject, `native` leaves selectors
alone), batch parity (`native` stays calibration-free, `ui` rides the
batch-wide calibration, selector rejects at parse time), Android parser
parity, and two new rows in `TapValidationParityTests` so the rejection is
pinned to fire identically on `tap`, `long-press` and `ios tap`.

Live, iPhone 17 / iOS 27.0, rotated to landscape-right. Two elements, each
compared against the selector path for the same element — that path is the
reference transform, so equality is the property under test:

| target | outline (ui) centre | `native` dispatch | `--coordinate-space ui` | selector dispatch |
|---|---|---|---|---|
| tab-bar item | (156.5, 46.0) | (156.5, 46.0) | **(356.0, 156.5)** | **(356.0, 156.5)** |
| nav-bar button | (701.0, 46.0) | (701.0, 46.0) | **(356.0, 701.0)** | **(356.0, 701.0)** |

One caveat stated plainly: I confirmed the dispatched point, not the visual
hit. On the simulator I had, the pre-existing selector path also produced no
visible change, so HID delivery there was not a usable signal — and since the
new path and the reference path dispatch the identical point, the transform is
what this change is responsible for, and that is what the table pins.

## Commits

Three, each building and passing on its own:

1. `fix(tap):` — the flag, the validation rule, both iOS dispatch paths, tests
2. `fix(android):` — the parity no-op on `android tap`
3. `docs(tap):` — skill, `--help` example, changelog

The bundled skill described this contract by symptom rather than by verb: its
Act table listed `tap -x/-y` as a last resort without naming the space the
numbers were in, and the orientation-advisory row scoped its remedy to
`swipe` / `touch`, which is where the flag used to stop. Both now name `tap`.

Refs #142


onevcat:
@onevclaw Review this PR and suggest a merge decision.

onevclaw:
审查结论：建议合并。

这组改动让 `tap` 的显式坐标支持 `--coordinate-space`，并与既有 `swipe` / `touch` 的校准路径保持一致；iOS 批处理与 Android 的参数兼容性也已覆盖。四项 CI 均通过，审查中未发现阻塞问题。

一个非阻塞的后续建议：README 总览仍只列出 `swipe` / `touch`，可以在后续补上 `tap`，让入口文档和能力表一致。

> 🧵 mh_1789534708898_poll-one · 🐾 main → main
> 🪝 gh:lycorp-jp:sim-use:issue:144:main:m202609
> 🆔 1b3ae599-d770-4d49-b734-aaf33fd3bf37

onevcat:
@iXerol LGTM. Thank you for this fix!

